PR #3, opened 2022-06-09
Fix docs

MatteoGheza:
I was reading the documentation, when I found some bugs, so I tried to fix them.


micheleriva:
Thank you!

PR #4, opened 2022-06-17
chore(lyra): small refactor of variable check

thomscoder:
I was reading through the code, I really loved this one, and it got me wondering if it is something I'm not getting or if we could just skip a condition and run the check against the returned Map size.

I take this opportunity to tell that I think the ./package.json -> benchmark script contains an outdated path to benchmarks after the build pipeline.


micheleriva:
Thank you @thomscoder!
Originally, Lyra was using objects instead of maps, and during the first refactor I forgot to remove those checks.

I'm going to merge this, thanks again

thomscoder:
Thanks for the explanation :) Best of luck for your next talks!

Issue #1, opened 2022-06-07
Add exact match


micheleriva:
**Is your feature request related to a problem? Please describe.**
Right now, Lyra only implements prefix search.

**Describe the solution you'd like**
Lyra needs to implement an exact search. I.E.:

```js
await lyra.search({
  term: "now",
  limit: 10,
  offset: 0,
  exact: true
});
```

Lyra should match:

```js
{
  txt: "now is time"
}
```

but not:

```js
{
  txt: "nowadays"
}
```

PR #8, opened 2022-07-12
Exact match

mateonunez:
This PR implements the #1 using a parameter that determinates if the search actions requires an exact match or not.

The `find` method of Trie nodes now accept an object as a single parameter. The type has been added: `FindParams`.

The solution found is to insert a check in the recursive function `findAllWords`, if the search actions requires an **exact** match and the _word_ given by the last node doesn't match with the _prefix_, returns the initial output.

I wrote the tests for this, I'm considering add more tests if necessary.

### DIFF
--- a/packages/lyra/src/lyra.ts
+++ b/packages/lyra/src/lyra.ts
@@ -25,6 +25,7 @@ export type SearchParams = {
   properties?: "*" | string[];
   limit?: number;
   offset?: number;
+  exact?: boolean | undefined;
 };
 
 type LyraIndex = Map<string, Trie>;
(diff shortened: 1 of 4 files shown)


packages/lyra/src/lyra.ts
```diff
@@ -25,6 +25,7 @@ export type SearchParams = {
   properties?: "*" | string[];
   limit?: number;
   offset?: number;
+  exact?: boolean | undefined;
```
micheleriva, line 28: We could get rid of `| undefined`, if the value is not defined will be interpreted as `false` by default

micheleriva (changes requested):

micheleriva:
Hi @mateonunez!
Thank you so much for your contribution. I left a couple of comments 🙂

micheleriva (approved):

micheleriva:
Awesome, LGTM!

Issue #2, opened 2022-06-07
Typo tolerance


micheleriva:
**Is your feature request related to a problem?**
As for now, Lyra does not implement typo tolerance during the search.

That means that if I misspell something like `seaorse` instead of `seahorse`, I don't get the expected results.

**Describe the solution you'd like**
Lyra should implement typo-tolerance using the Levenshtein algorithm or similar while traversing the trie.
Users should be able to disable the typo-tolerance feature as follows:

```js
await lyra.search({
  term: "seaorse",
  limit: 5,
  offset: 2,
  exact: true,
  fixTypos: false
})
```

`fixTypos` should be enabled by default.

PR #11, opened 2022-07-13
Tolerance

mateonunez:
This PR is a proposal to resolve the problem of #2.

The changes don't implement a `boolean` to manage the typo errors. Instead of that, a new parameter has been added to the `search` method.

As the documentation shows: tolerance is the **maximum** distance between the searched _term_ and a _word_ inside a document.

I've added a few tests for this implementation.

### Coverage

**Without Tolerance implementation**:

| File          | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s |
| ------------- | ------- | -------- | ------- | ------- | ----------------- |
| All files     | 85.6    | 65.38    | 89.47   | 85.82   |                   |
| ...refix-tree | 95.18   | 78.57    | 100     | 97.5    |                   |
| trie.ts       | 94.02   | 78.57    | 100     | 96.87   | 104,129           |

**With Tolerance implementation**:
| File          | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s |
| ------------- | ------- | -------- | ------- | ------- | ----------------- |
| All files     | 86.44   | 67.41    | 90      | 87.14   |                   |
| src           | 82.43   | 58.18    | 84.61   | 82.38   |                   |
| ...shtein.ts  | 91.66   | 60       | 100     | 100     | 2-3               |
| ...refix-tree | 95.55   | 82.35    | 100     | 97.7    |                   |
| trie.ts       | 94.59   | 82.35    | 100     | 97.18   | 115,140           |

### Time elapsed

**Without Tolerance implementation**:
```js
{
  elapsed: '135μs',
  hits: [
    {
      id: 'ECbOj5pixIOJc9KfQpwss',
      quote: 'Be yourself; everyone else is already taken.',
      author: 'Oscar Wilde',
      isFavorite: false,
      rating: 4
    },
    {
      id: 'jMfGCEX1zf-TX0Tc5-e-8',
      quote: 'So many books, so little time.',
      author: 'Frank Zappa',
      isFavorite: true,
      rating: 5
    }
  ],
  count: 2
}
```

**With Tolerance implementation**:
```js
{
  elapsed: '147μs',
  hits: [
    {
      id: 'FFU-21lklMCkviNQgSPCc',
      quote: 'Be yourself; everyone else is already taken.',
      author: 'Oscar Wilde',
      isFavorite: false,
      rating: 4
    },
    {
      id: '4FAnVKAP2wygWEbnZOUhT',
      quote: 'So many books, so little time.',
      author: 'Frank Zappa',
      isFavorite: true,
      rating: 5
    }
  ],
  count: 2
}
```

### DIFF
--- a/packages/lyra/src/prefix-tree/trie.ts
+++ b/packages/lyra/src/prefix-tree/trie.ts
@@ -66,6 +66,7 @@ export class Trie {
     findAllWords(node, output);
 
     function findAllWords(_node: TrieNode, _output: FindResult) {
+      let distance = 0;
       if (_node.end) {
         const [word, docIDs] = _node.getWord();
 
@@ -74,9 +75,13 @@ export class Trie {
         }
 
         if (!(word in _output)) {
-          const distance = levenshtein(term, word);
+          if (tolerance) {
+            distance = levenshtein(term, word);
 
-          if (distance <= (tolerance || (tolerance !== 0 && word.length))) {
+            if (distance <= tolerance) {
+              _output[word] = new Set();
+            }
+          } else {
             _output[word] = new Set();
           }
         }
--- a/packages/lyra/src/prefix-tree/trie.ts
+++ b/packages/lyra/src/prefix-tree/trie.ts
@@ -1,8 +1,10 @@
 import { TrieNode } from "./node";
+import { levenshtein } from "../levenshtein";
 
 export type FindParams = {
   term: string;
   exact?: boolean;
+  tolerance?: number;
 };
 
 export type FindResult = {
@@ -49,14 +51,14 @@ export class Trie {
     return node.end;
   }
 
-  find({ term, exact }: FindParams): FindResult {
+  find({ term, exact, tolerance }: FindParams): FindResult {
     let node = this.root;
     const output: FindResult = {};
 
     for (const char of term) {
       if (node?.children?.has(char)) {
         node = node.children.get(char)!;
-      } else {
+      } else if (!tolerance) {
         return output;
       }
     }
@@ -72,12 +74,16 @@ export class Trie {
         }
 
         if (!(word in _output)) {
-          _output[word] = new Set();
+          const distance = levenshtein(term, word);
+
+          if (distance <= (tolerance || (tolerance !== 0 && word.length))) {
+            _output[word] = new Set();
+          }
         }
 
         if (docIDs?.size) {
           for (const doc of docIDs) {
-            _output[word].add(doc);
+            _output[word] && _output[word].add(doc);
           }
         }
       }
--- a/packages/lyra/src/prefix-tree/trie.ts
+++ b/packages/lyra/src/prefix-tree/trie.ts
@@ -66,7 +66,6 @@ export class Trie {
     findAllWords(node, output);
 
     function findAllWords(_node: TrieNode, _output: FindResult) {
-      let distance = 0;
       if (_node.end) {
         const [word, docIDs] = _node.getWord();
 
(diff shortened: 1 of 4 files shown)


micheleriva:
Thank you @mateonunez, this is gold. I'll review the code tomorrow morning and let you know how we want to proceed.

I actually planned to introduce a `tolerance`-like property in one of the next releases of Lyra, but if we get it merged sooner, it's better.

mateonunez:
Sounds great @micheleriva! Can't wait for the new release and the new features. 👌

packages/lyra/src/prefix-tree/trie.ts
```diff
@@ -72,12 +75,24 @@ export class Trie {
         }
 
         if (!(word in _output)) {
-          _output[word] = new Set();
+          if (tolerance) {
+            // if the tolerance is set, we need to calculate the distance using levenshtein algorithm
+            distance = levenshtein(term, word);
```
micheleriva, line 80: This looks perfect, I would just add a simple control like `if (word.length <= tolerance)` to avoid computing the edit distance on large strings, which will be ignored anyway. What do you think?

packages/lyra/tests/lyra.test.ts
```diff
@@ -305,4 +305,61 @@ describe("lyra", () => {
     );
     expect((exactSearch.hits[0] as any).author).toBe("Oscar Wilde");
   });
+
+  it("Shouldn't be tolerant", async () => {
```
micheleriva, line 309: I would call this test `it("Shouldn't tolerate typos")`

packages/lyra/tests/lyra.test.ts
```diff
@@ -305,4 +305,61 @@ describe("lyra", () => {
...
+
+    const search = await db.search({
+      term: "seahrse",
+      tolerance: 0,
+    });
+
+    expect(search.count).toBe(0);
+  });
+
+  it("Should be tolerant", async () => {
```
micheleriva, line 331: I would call this test `it("Should tolerate typos")`

micheleriva:
@mateonunez this PR looks really good! I just left a few minor comments, it would be awesome if you could address them.

In a future release, as an improvement, we should aim at running the Levenshtein algorithm against the tree instead of words.

Let me give you an example:

* If you're searching for `"hello"`, but you mistype and write `"hllo"` instead, the current algorithm will work just fine.
* If you're searching for `"hello"`, but you mistype and write `"yellow"` (with a tolerance of `2`), this won't work, as we discard words with `y` as a prefix.

This is something to consider for future releases, as for now, I am happy with the current PR

packages/lyra/src/prefix-tree/trie.ts
mateonunez: I was thinking about it, I've added a check in the first commit that checks the **word.lenght**, but I wasn't completely convinced.

We can do something like this:

```js
if (tolerance) {
  // computing the difference of letters between the term and the word
  const difference = Math.abs(term.length - word.length);

  if (difference <= tolerance) {
    // if the tolerance is set, we need to calculate the distance using levenshtein algorithm
    distance = levenshtein(term, word);
...
```

Computing the difference of letters between the **word** and the **term** we can avoid on large strings.

It could be a good solution in your opinion?

micheleriva: Yes, let's do this!

mateonunez:
I got the point about the next Levenshtein algorithm implementation. It might be fun to work on it.

micheleriva (approved):

micheleriva:
LGTM

PR #19, opened 2022-07-16
fix(benchmark start command): bennchmark start command and files import

ilteoood:
With this PR I would like to improve the benchmark package, that currently isn't working.

Even after this, the `title.csv` required by `prefix-search-movies` is still missing.


micheleriva:
Thanks! I need to include a bash/js command to download and normalize one of the following files for the benchmark (I don't want to commit such a large file): https://datasets.imdbws.com

micheleriva (approved):

PR #21, opened 2022-07-17
Typed hits property in search method

DanieleFedeli:
This PR aims to avoid the assertion of a type when handling the result object.

### DIFF
--- a/packages/lyra/src/lyra.ts
+++ b/packages/lyra/src/lyra.ts
@@ -21,7 +21,7 @@ export type LyraProperties<
   defaultLanguage?: Language;
 };
 
-export type LyraDocs = Map<string, object>;
+export type LyraDocs<TDoc> = Map<string, TDoc>;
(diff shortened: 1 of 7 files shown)


packages/lyra/src/lyra.ts
```diff
@@ -129,7 +135,7 @@ export class Lyra<TSchema extends PropertiesSchema = PropertiesSchema> {
               break;
             }
 
-            const fullDoc = this.docs.get(id);
+            const fullDoc = this.docs.get(id)!;
```
DanieleFedeli, line 138: @micheleriva Is there a better way to do that? I hate asserting non-null

micheleriva: mmm afaik this is the only way... if there's a guarantee that the value exists, it's not that bad

micheleriva (approved):

Issue #15, opened 2022-07-16
Extend query parameters by adding query clauses


micheleriva:
**Is your feature request related to a problem? Please describe.**
As for now, Lyra is capable of indexing documents with searchable and non-searchable fields.

For instance, given the following schema, we index the following fields:

```js
import { lyra } from '@nearfom/lyra';

const movieDB = new Lyra({
  schema: {
    // searchable fields
    title: 'string',
    director: 'string',
    plot: 'string',

    // non searchable
    year: 'number',
    isFavorite: 'boolean'
  }
});
```

Even though numbers and booleans are non-searchable fields, we should start using them for performing queries using the `where` keyword. An example could be:

```js
const result = await movieDB.search({
  term: 'love',
  limit: 10,
  offset: 5,
  where: {
    year: { '>=': 1990 },
    isFavorite: true
  }
});
```

as a first iteration, we could go using `AND` only (so, in the above example, `WHERE year >= 1990 AND isFavorite = true`). In the future, we might want to support `AND`, `OR`, `CONTAINS`, etc.

DanieleFedeli:
How can this be done? 
Do we first retrieve the documents and then apply the conditions or we are also indexing non-searchable fields?

micheleriva:
@DanieleFedeli I think we could store non-searchable parameters in reverse indexes and then access these properties in constant time during the search. I am currently working on the stemming issue, but I can try to illustrate my proposal with a simple, non-working, demonstrative PR

micheleriva:
TL;DR

![6i3b18 (1)](https://user-images.githubusercontent.com/14977595/179469495-61264aad-be3d-47be-baed-d7889a743f89.jpeg)

DanieleFedeli:
For now I am using these data structures for indexing non-searchable fields:
- Boolean: HashMap -> Key: `${PropName}_${booleanValue}` Value: Set which contains ids
- Number: HashMap -> Key: `${PropName}` value: another HashMap (Key -> numeric value, Value -> Set which contains ids)

Imho the boolean approach is ok but I personally don't like the numeric one. For exact numbers, it works well with just two reads in O(1) but when we have to do a range query it is a bit hacky. 

Do you have any suggestions on that?

DanieleFedeli:
My PR is almost ready, I will try to do some benchmark

micheleriva:
[...]

PR #36, opened 2022-07-20
Support query clauses

DanieleFedeli:
Targets #15 
Functionality is ready.
I need some benchmarks and to resolve conflicts.


DanieleFedeli:
@micheleriva  Benchmarks:

Screenshot 2022-07-20 alle 21 14 16

Honestly, I am not happy with the range of query results.  Right now I do a linear search over keys. I will try something non-linear.

packages/lyra/src/lyra.ts
```diff
@@ -97,8 +126,13 @@ export class Lyra<TSchema extends PropertiesSchema = PropertiesSchema> {
 
       if (isNested) {
         this.buildIndex(schema[prop] as TSchema, `${propName}.`);
-      } else {
+      } else if (propValue === "string") {
```
anonrig, line 129: Numeric comparisons are faster than string based comparisons. What about having a constant of ValueType.STRING with a value of number?

packages/lyra/src/utils.ts
```diff
@@ -29,3 +31,28 @@ export function getNanosecondsTime(): bigint {
...
+  }
+  return insersection;
+}
+
+export function setSubtraction<T = unknown>(
+  from: Nullable<Set<T>> = new Set<T>(),
+  toRemove: Set<T> = new Set<T>()
+): void {
+  for (const element of toRemove) {
+    from?.delete(element);
```
anonrig, line 56: from has a default value, I assume `?` operator is unnecessary.

packages/lyra/src/lyra.ts
```diff
@@ -187,6 +244,62 @@ export class Lyra<TSchema extends PropertiesSchema = PropertiesSchema> {
...
+    schema: TSchema,
+    prefix = ""
+  ): GetDocumentFromWhereParams {
+    const boolIndices: string[] = [];
+    const numericIndices: string[] = [];
+    if (!where) return { boolIndices, numericIndices };
+
+    for (const key of Object.keys(where)) {
+      const propName = `${prefix}${String(key)}`;
+      const propValue = (where as any)[key];
```
anonrig, line 258:
```suggestion
      const propValue = where[key as keyof where];
```

packages/lyra/src/lyra.ts
```diff
@@ -120,36 +160,53 @@ export class Lyra<TSchema extends PropertiesSchema = PropertiesSchema> {
...
+          exact: exact,
+          where: where,
         });
 
+        // To avoid duplicates we need a subtraction between added document
+        // And retrieved ones.
+        // We check if this is the first
+        // We use non-null assertion because setIntersection will return null only if both set are null
+        // We are sure that documentsIDs is not null, so we are safew
+        documentIDs = setIntersection(
```
anonrig, line 183: Can you benchmark this particular line? I remember re-initializing and changing the type of a variable, which affects the JIT performance.

anonrig (commented):
Even though I didn't contribute to Lyra at all, I wanted to see and contribute by leaving a review here. I hope it'll help with the performance regression.

Great contribution btw. Thanks! 💯 🎉

DanieleFedeli:
Thank you @anonrig 
I will work on that soon :clinking_glasses:

packages/lyra/src/utils.ts
DanieleFedeli: You right, I will remove that

packages/lyra/src/lyra.ts
DanieleFedeli: Ok I will try to do a benchmark with a constant variable, let's see the difference

packages/lyra/src/lyra.ts
DanieleFedeli: I am not getting what you are suggesting here, can you please elaborate?

DanieleFedeli:
Managed to get a better benchmark by changing the index data structure.

Screenshot 2022-07-22 alle 19 43 02

I will solve the conflicts, adjust the document deletion and I'm done.

DanieleFedeli:
Hi @micheleriva, Is there something wrong with this PR? I can fix it if that is the case

micheleriva:
Hey @DanieleFedeli, I'm putting this PR on hold until we merge a major change from the `simple-structures` branch

DanieleFedeli:
Ok, please write a comment when we can merge this so I will resolve conflicts

packages/lyra/src/lyra.ts
```diff
@@ -120,36 +163,54 @@ export class Lyra<TSchema extends PropertiesSchema = PropertiesSchema> {
...
+        // And retrieved ones.
+        // We check if this is the first
+        // We use non-null assertion because setIntersection will return null only if both set are null
+        // We are sure that documentsIDs is not null, so we are safew
+        const documentIDs = setIntersection(
+          documentIDsFromText,
+          documentIdsFromNonSearchable
+        )!;
+
+        setSubtraction(documentIDs, documentAdded);
```
DanieleFedeli, line 191: @micheleriva even if this PR is not getting merged, I think that the set difference solves #48

packages/lyra/src/lyra.ts
```diff
@@ -4,12 +4,28 @@ import * as fastq from "fastq";
...
+  ResolveSchema,
+  SearchProperties,
+  WhereParams,
+} from "./types";
+import { SortedQueue } from "./sorted-queue";
+import { QueueNode } from "./sorted-queue/node";
 
 export type PropertyType = "string" | "number" | "boolean";
 
+export const allowedNumericComparison = [">=", ">", "=", "<", "<="] as const;
```
stearm, line 27: Can I ask why a symbolic strategy was chosen instead of a literal one? So, for instance: `gte`, `gt`, `eq`, `lt`, `lte` etc.

If clauses like "includes," "between," and so on are introduced in the future, everything will be more consistent. What are your thoughts?

stearm, line 27: Why not `!=`?

DanieleFedeli: @micheleriva Suggested that in the issue. I prefer the symbolic one since I am mostly used to it, but we can start a discussion on it.

DanieleFedeli: We started with the simplest one, later we can implement things like INCLUDES, LIKE, OR and !=.

I excluded that because I think that is harder to implement now in this PR since I need to do a linear search. We can do that in the future.

DanieleFedeli:
@micheleriva May I start developing the plugin?

micheleriva:
@DanieleFedeli we're still working on the hooks system, I'd recommend waiting for it as it would simplify the work by a lot 🙏

DanieleFedeli:
@micheleriva I have been busy this period, I did not follow Lyra. 

Are hooks integrated?

Issue #40, opened 2022-07-21
Deleting one document also causes other documents to disappear from searches


lucaong:
**Describe the bug**
The bug occurs in `lyra-0.0.1-beta-13`.

Sorry for opening several bugs, I was trying to compare Lyra with [MiniSearch](https://github.com/lucaong/minisearch) and ran into these issues.

**To Reproduce**

```javascript
const db = new Lyra({
  schema: {
    txt: "string",
  }
})

await db.insert({ txt: 'abc' })
//=> { id: 'J-LiSKu45O4d1phnVWvkc' }
await db.insert({ txt: 'abc' })
//=> { id: 'yx-s9o0-8I4uWcPsMyOWL' }
await db.insert({ txt: 'abcd' })
{ id: 'XoKJ6uWhRkftAt06UsdKo' }

await db.search({ term: 'abc', exact: true })
/* Returns what expected:
{
  elapsed: '814μs',
  hits: [
    { id: 'J-LiSKu45O4d1phnVWvkc', txt: 'abc' },
    { id: 'yx-s9o0-8I4uWcPsMyOWL', txt: 'abc' },
    undefined,
    undefined,
    undefined,
    undefined,
    undefined,
    undefined,
    undefined,
    undefined
  ],
  count: 2
}
*/

// Delete one of the two documents containing term `"abc"`
db.delete('yx-s9o0-8I4uWcPsMyOWL')

await db.search({ term: 'abc', exact: true })
/* Returns no results, even if one doc with "abc" should still be in the index:
{
  elapsed: '770μs',
  hits: [
    undefined, undefined,
    undefined, undefined,
    undefined, undefined,
    undefined, undefined,
    undefined, undefined
  ],
  count: 0
}
*/
```

**Expected behavior**

Deleting one document should have no effect on other documents.

lucaong:
I don't think this is fixed by #41 . The test in that PR has an issue that leads to a false positive.

lucaong:
Check my comment on #45, as it fixes the issue but leaves dead code and possible performance concerns.

Issue #38, opened 2022-07-21
Typo tolerance misses results


lucaong:
**Describe the bug**

The bug occurs in `lyra-0.0.1-beta-13`.

Typo tolerant searches miss expected results.

**To Reproduce**

```javascript
const db = new Lyra({
  schema: {
    txt: "string",
  },
  stemming: false // Disabling stemming to avoid confounding the results
})

// Insert "stelle", and other words that are all within edit distance 2
await db.insert({ txt: 'stelle' })
await db.insert({ txt: 'stele' })
await db.insert({ txt: 'snelle' })
await db.insert({ txt: 'stellle' })
await db.insert({ txt: 'scelte' })

await db.search({ term: 'stelle', tolerance: 2 })
/* returns:
{
  elapsed: '1ms',
  hits: [
    { id: 'cuD10UKGVzBmEQz30dQKI', txt: 'stelle' },
    { id: 'DqXro5_9z2wiPQlTmK91n', txt: 'stellle' },
    undefined,
    undefined,
    undefined,
    undefined,
    undefined,
    undefined,
    undefined,
    undefined
  ],
  count: 2
}
*/
```

**Expected behavior**

All 5 documents should be returned, since they all contain a term within edit distance 2 from the query.

micheleriva:
Looks like a bug, thank you for opening this!

tlahav:
Hey, I am looking for projects to contribute to and I found this issue in https://goodfirstissue.dev/. I looked at the code a bit and I feel like either I dont fully understand the feature or its implementation is a bit off. For example, the documentation ([here](https://nearform.github.io/lyra/guide/search.html#tolerance)) shows that searching for 'Cris' returns 1 result. But, with the same tolerance, searching for 'Bhris' should also return that same result, no? 'Bhris' doesnt work because the 'B' starts the find function down the 'Burton' subtree. Should 'Bhris' also work in this case? 

As for the original steps to reproduce, I think the issue is the fact that the original tree search is incompletable with the tolerance feature(?).  Basically, the current implementation uses the current subtree and checks for matches that can be made within that subtree, but that ignores all the other valid options. I believe i improved that but I feel like i still do not have a good grasp on the requirements here. Do check out: https://github.com/nearform/lyra/pull/47 

Thanks

micheleriva:
Hey @tlahav , you're correct, thank you for opening a PR 🙏

lucaong:
[...]

PR #41, opened 2022-07-22
Add exact param in remove doc by word method

### LINKED ISSUE #39: Deleted documents still appear in searches
**Describe the bug**

In some cases, deleted documents still appear in later searches

**To Reproduce**

The bug seems to depend on the specific content of the index. For example, start with a setup identical to #38 :

```javascript
const db = new Lyra({
  schema: {
    txt: "string",
  },
  stemming: false // Disabling stemming to avoid confounding the results
})

await db.insert({ txt: 'stelle' })
//=> { id: '3A5o53WhYbi6o1i2ZgO7B' }
await db.insert({ txt: 'stele' })
//=> { id: 'DDWCn_B1z5SF9TRr8yIOb' }
await db.insert({ txt: 'snelle' })
//=> { id: 'XF3RgLkQEMY0OlFn95Nl_' }
await db.insert({ txt: 'stellle' })
//=> { id: 'UAUueC4rOGfnFhky7oEMm' }
await db.insert({ txt: 'scelte' })
//=> { id: '04fWgLHV3zA9GQHm8mFay' }

await db.search({ term: 'stelle', tolerance: 2 })
/* It returns two results, which is incorrect but not the point of this issue (see #38 for that):
{
  elapsed: '624μs',
  hits: [
    { id: '3A5o53WhYbi6o1i2ZgO7B', txt: 'stelle' },
    { id: 'UAUueC4rOGfnFhky7oEMm', txt: 'stellle' },
    undefined,
    undefined,
    undefined,
    undefined,
    undefined,
    undefined,
    undefined,
    undefined
  ],
  count: 2
}
*/

// Now delete one of the documents that were returned by the previous query:
await db.delete('UAUueC4rOGfnFhky7oEMm')
#=> true

await db.search({ term: 'stelle', tolerance: 2 })
/* Still returns two results:
{
  elapsed: '624μs',
  hits: [
    { id: '3A5o53WhYbi6o1i2ZgO7B', txt: 'stelle' },
    { id: 'UAUueC4rOGfnFhky7oEMm' },
    undefined,
    undefined,
    undefined,
    undefined,
    undefined,
    undefined,
    undefined,
    undefined
  ],
  count: 2
}
*/
```

Note that the last search results include the deleted document, although with no fields apart from the `id`.

**Expected behavior**

Deleted documents should not be returned by any search performed after the deletion.

mateonunez:
This PR fixes #39 and #40.

I've implemented a new param in the `removeDocByWord` method used in `remove` method to prevent removing words that not exactly matches the document's word.

I also added 2 cases to the tests.

### DIFF
--- a/packages/lyra/src/prefix-tree/trie.ts
+++ b/packages/lyra/src/prefix-tree/trie.ts
@@ -107,14 +107,14 @@ export class Trie {
     return output;
   }
 
-  removeDocByWord(word: string, docID: string): boolean {
+  removeDocByWord(word: string, docID: string, exact = false): boolean {
     const root = this.root;
     if (!word) return false;
 
     function removeWord(node: TrieNode, _word: string, docID: string): boolean {
       const [nodeWord /**_docs*/] = node.getWord();
 
-      if (node.end && nodeWord === word) {
+      if (node.end || (exact && node.end && nodeWord === word)) {
         node.removeDoc(docID);
 
         if (node.children?.size) {
--- a/packages/lyra/tests/lyra.test.ts
+++ b/packages/lyra/tests/lyra.test.ts
@@ -195,6 +195,51 @@ describe("lyra", () => {
     expect(searchResult.hits[0].id).toBe(id2);
   });
 
+  it("Shouldn't returns deleted documents", async () => {
+    const db = new Lyra({
+      schema: {
+        txt: "string",
+      },
+      stemming: false,
+    });
+
+    await db.insert({ txt: "stelle" });
+    await db.insert({ txt: "stellle" });
+    await db.insert({ txt: "scelte" });
+
+    const search = await db.search({ term: "stelle" });
+
+    const id = search.hits[0].id;
+
+    await db.delete(id);
+
+    const serach2 = await db.search({ term: "stelle" });
+
+    expect(serach2.count).toBe(1);
+  });
+
+  it("Shouldn't affects other document when deleted one", async () => {
+    const db = new Lyra({
+      schema: {
+        txt: "string",
+      },
+      stemming: false,
+    });
+
+    await db.insert({ txt: "abc" });
+    await db.insert({ txt: "abc" });
+    await db.insert({ txt: "abcd" });
+
+    const search = await db.search({ term: "abc", exact: true });
+
+    const id = search.hits[0].id;
+    await db.delete(id);
+
+    const search2 = await db.search({ term: "abc" });
+
+    expect(search2.count).toBe(1);
+  });
+
   it("Should preserve identical docs after deletion", async () => {
     const db = new Lyra({
       schema: {


micheleriva (approved):

micheleriva:
This is perfect, thank you @mateonunez 🙏

packages/lyra/tests/lyra.test.ts
```diff
@@ -195,6 +195,51 @@ describe("lyra", () => {
...
+    await db.insert({ txt: "abc" });
+    await db.insert({ txt: "abc" });
+    await db.insert({ txt: "abcd" });
+
+    const search = await db.search({ term: "abc", exact: true });
+
+    const id = search.hits[0].id;
+    await db.delete(id);
+
+    const search2 = await db.search({ term: "abc" });
```
lucaong, line 238: This line should also use `exact: true`, otherwise the tests passes even if #40 is not fixed.

packages/lyra/src/prefix-tree/trie.ts
```diff
@@ -107,14 +107,14 @@ export class Trie {
...
-  removeDocByWord(word: string, docID: string): boolean {
+  removeDocByWord(word: string, docID: string, exact = false): boolean {
     const root = this.root;
     if (!word) return false;
 
     function removeWord(node: TrieNode, _word: string, docID: string): boolean {
       const [nodeWord /**_docs*/] = node.getWord();
 
-      if (node.end && nodeWord === word) {
+      if (node.end || (exact && node.end && nodeWord === word)) {
```
lucaong, line 117: I don't think this fixes #40 . At a glance, #40 is caused by setting `end` to `false` without checking if the term is associated to other documents.

packages/lyra/tests/lyra.test.ts
mateonunez: If you search for `abc` as an exact term, the result will be **0**, 'cause the only document available is the document that contains `txt: "abcd"`. 

image

This expected result could be more realistic, but in both cases, it works correctly

lucaong: But it shouldn't be 0, that's the bug in #40. It should be 1, because there is a second document containing `"abc"` that should **not** be deleted.

packages/lyra/src/prefix-tree/trie.ts
mateonunez: I don't think so, #40 was caused because the document was deleted only when the words were the same (implementing `exact` parameter we don't have that issue). 

I think #38 is caused for the reason you said. I'm working on it.

packages/lyra/tests/lyra.test.ts
mateonunez: Got the point. The issue should be opened again.

packages/lyra/src/prefix-tree/trie.ts
lucaong: #38 cannot be caused by this, because it occurs even if no document is deleted, so this code path is never called in the reproduction.

packages/lyra/tests/lyra.test.ts
```diff
@@ -195,6 +195,51 @@ describe("lyra", () => {
...
+
+    const search = await db.search({ term: "stelle" });
+
+    const id = search.hits[0].id;
+
+    await db.delete(id);
+
+    const serach2 = await db.search({ term: "stelle" });
+
+    expect(serach2.count).toBe(1);
```
lucaong, line 218: This assertion should check that the search hits do not include the deleted ID, not that the count is 1.

PR #43, opened 2022-07-23
Prevent exact match in ended node

### LINKED ISSUE #40: Deleting one document also causes other documents to disappear from searches

mateonunez:
This PR would fix #40.

The problem was caused because the `findAllWords` method considered only nodes with `end` as **true**. In this case, the Node would contain other children (with the exact word matched) Lyra will consider only the first node returning the wrong word and comparing it to the term.

I've recycled the `handleOutput` method, that cycles the `docIDs` and adds to the output the document id.

Fixed the false positive.

```bash
> mun@mateounez MINGW64 ~/Development/forks/lyra (main)
$ git checkout main && pnpm just:testing
Already on 'main'
M       package.json
Your branch is up to date with 'origin/main'.

> lyra@1.3.0 just:testing C:\Users\nunmat\Development\forks\lyra
> ts-node ./packages/lyra/just-testing.ts

{ elapsed: '13μs', hits: [], count: 0 }

> mun@mateonunez MINGW64 ~/Development/forks/lyra (main)
$ git checkout fix/trie-exact-match && pnpm just:testing
Switched to branch 'fix/trie-exact-match'
M       package.json
Your branch is up to date with 'origin/fix/trie-exact-match'.

> lyra@1.3.0 just:testing C:\Users\nunmat\Development\forks\lyra
> ts-node ./packages/lyra/just-testing.ts

{
  elapsed: '17μs',
  hits: [
    { id: 'yZFVgKBfe5dqXekBUmuQg', txt: 'abc' },
    { id: '8htoGu6BrItYk0u6c8UgQ', txt: 'abc' }
  ],
  count: 2
}
```

### DIFF
--- a/packages/lyra/src/prefix-tree/trie.ts
+++ b/packages/lyra/src/prefix-tree/trie.ts
@@ -63,6 +63,18 @@ export class Trie {
       }
     }
 
+    const [word, docIDs] = node.getWord();
+
+    if (exact && word === term) {
+      output[word] = new Set();
+
+      for (const doc of docIDs) {
+        output[word].add(doc);
+      }
+
+      return output;
+    }
+
     findAllWords(node, output);
 
     function findAllWords(_node: TrieNode, _output: FindResult) {
--- a/packages/lyra/src/prefix-tree/trie.ts
+++ b/packages/lyra/src/prefix-tree/trie.ts
@@ -63,27 +63,34 @@ export class Trie {
       }
     }
 
+    // avoid the recursive call if the word is the same as the term
     const [word, docIDs] = node.getWord();
 
     if (exact && word === term) {
       output[word] = new Set();
 
-      for (const doc of docIDs) {
-        output[word].add(doc);
-      }
+      handleOutput(word, docIDs, output);
 
       return output;
     }
 
     findAllWords(node, output);
 
+    function handleOutput(
+      word: string,
+      docIDs: Set<string>,
+      _output: FindResult
+    ): void {
+      if (docIDs.size) {
+        for (const doc of docIDs) {
+          _output[word] && _output[word].add(doc);
+        }
+      }
+    }
+
     function findAllWords(_node: TrieNode, _output: FindResult) {
       if (_node.end) {
         const [word, docIDs] = _node.getWord();
-
-        if (exact && word !== term) {
-          return output;
-        }
 
         if (!(word in _output)) {
           if (tolerance) {
@@ -103,12 +110,7 @@ export class Trie {
           }
         }
 
-        if (docIDs?.size) {
-          for (const doc of docIDs) {
-            // check if _output[word] exists and then add the doc to it
-            _output[word] && _output[word].add(doc);
-          }
-        }
+        handleOutput(word, docIDs, _output);
       }
 
       for (const childNode of _node.children?.values() ?? []) {
--- a/packages/lyra/tests/lyra.test.ts
+++ b/packages/lyra/tests/lyra.test.ts
@@ -229,8 +229,10 @@ describe("lyra", () => {
     await db.insert({ txt: "abc" });
     await db.insert({ txt: "abc" });
     await db.insert({ txt: "abcd" });
+    await db.insert({ txt: "abc" });
+
     await db.insert({ txt: "abcd" });
-    await db.insert({ txt: "abcd" });
+    await db.insert({ txt: "abc" });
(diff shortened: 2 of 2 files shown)


packages/lyra/src/prefix-tree/trie.ts
```diff
@@ -63,16 +63,35 @@ export class Trie {
...
 
+    // avoid the recursive call if the word is the same as the term
+    const [word, docIDs] = node.getWord();
+
+    if (exact && word === term) {
+      output[word] = new Set();
+
+      handleOutput(word, docIDs, output);
+
+      return output;
```
lucaong, line 74: This looks problematic: early return here means that no prefix or fuzzy search is performed. Note that, since there is no check for the node being a terminal one, this happens even when there is no exact match.

For example, say you have only the single word `star` in the whole index. Imagine searching for `sta` allowing for prefix match. You expect to find the documents associated to `star` due to prefix match, instead the logic finds `sta` (without checking that is not a terminal node), and returns no result because there is no document associated to it.

lucaong (commented):
Unfortunately, I think this breaks more than it fixes (see comments).

The cause of #40 is not in the find logic. It’s in the document deletion logic. As far as I can see, it’s because `removeDocByWord` sets `node.end = false` (effectively deleting the word) without checking if there were other documents associated to the same word.

packages/lyra/src/prefix-tree/trie.ts
```diff
@@ -63,16 +63,35 @@ export class Trie {
...
     findAllWords(node, output);
 
+    function handleOutput(
+      word: string,
+      docIDs: Set<string>,
+      _output: FindResult
+    ): void {
+      if (docIDs.size) {
+        for (const doc of docIDs) {
+          _output[word] && _output[word].add(doc);
```
thomscoder, line 86: Heyy @mateonunez 👋🏽
I'm wondering if changing this in

```javascript
const val = _output[word];
if (docIDs.size && val) {
 for (const doc of docIDs) {
  val.add(doc);
 }
}
```
would prevent potential overhead on falsy `_output[word]`. 
- It seems that `_output[word]` is not changing (will stop accessing _output all the time).
- therefore we could enter the loop only if `_output[word]` is truthy. 

> Obviously the overhead/perfs depend on the size of the Set
<hr>
Anyway, thanks in advance for your feedback, eventually 😄

packages/lyra/src/prefix-tree/trie.ts
mateonunez: Thank you Luca for checking these things for me, I will try to find a final solution to the problem by following your great suggestions!

packages/lyra/src/prefix-tree/trie.ts
mateonunez: I think the PR needs to make other logic, I will close this one trying to solve it definitively.

PR #45, opened 2022-07-24
Remove by word should consider other doc ids

### LINKED ISSUE #40: Deleting one document also causes other documents to disappear from searches

mateonunez:
This PR would fix #40.

Thanks a lot to @lucaong for supporting me in this bug.

The `removeByWord` now considers if the **docID** exists in _docIDs_ Set, then set the `node.end` as `false`. 
The `exact` param will stay in the function for future implementations.

### DIFF
--- a/packages/lyra/src/prefix-tree/trie.ts
+++ b/packages/lyra/src/prefix-tree/trie.ts
@@ -112,12 +112,12 @@ export class Trie {
     if (!word) return false;
 
     function removeWord(node: TrieNode, _word: string, docID: string): boolean {
-      const [nodeWord /**_docs*/] = node.getWord();
+      const [nodeWord, docIDs] = node.getWord();
 
       if (node.end || (exact && node.end && nodeWord === word)) {
         node.removeDoc(docID);
 
-        if (node.children?.size) {
+        if (node.children?.size && docIDs.has(docID)) {
           node.end = false;
         }
 
--- a/packages/lyra/tests/lyra.test.ts
+++ b/packages/lyra/tests/lyra.test.ts
@@ -235,8 +235,9 @@ describe("lyra", () => {
     const id = search.hits[0].id;
     await db.delete(id);
 
-    const search2 = await db.search({ term: "abc" });
+    const search2 = await db.search({ term: "abc", exact: true });
 
+    expect(search2.hits.every(({ id: docID }) => docID !== id)).toBeTruthy();
     expect(search2.count).toBe(1);
   });
 


thomscoder:
LGTM 💯 Bug #40 seems to be fixed.

micheleriva:
Thank you @lucaong, @thomscoder, and @mateonunez for working on this one

packages/lyra/src/prefix-tree/trie.ts
```diff
@@ -112,12 +112,12 @@ export class Trie {
...
 
     function removeWord(node: TrieNode, _word: string, docID: string): boolean {
-      const [nodeWord /**_docs*/] = node.getWord();
+      const [nodeWord, docIDs] = node.getWord();
 
       if (node.end || (exact && node.end && nodeWord === word)) {
         node.removeDoc(docID);
 
-        if (node.children?.size) {
+        if (node.children?.size && docIDs.has(docID)) {
```
lucaong, line 120: This fixes the issue, but for the wrong reason :)

`docIDs.has(docID)` will always return `false` here, because in the previous line `node.removeDoc(docID)` removed the document from the set. As a result, `node.end = false` is never executed, so the whole `if` block here could be completely removed.

Never setting `node.end = false` on one hand fixes the issue, and does not cause any correctness problem. It just leaves a term in the trie, even when all documents associated to it were removed. This is fine, but will have performance impacts on large tries if a lot of documents are deleted: even when the set of documents associated to a term is empty, the term is never removed, so it is still considered in prefix and fuzzy search.

My recommendation: change the `if` condition to something like `node.children?.size && docIDs.size === 0`. That means: if after deleting this one document there are no more documents in the set pointed by this word, remove the word from the trie by making the node non-terminal.

Note that this still leaves one case uncovered, namely when the node its terminal but has no children. In that case, in order to really delete the term from the trie, one has to backtrack to the closest branching, and adjust the nodes to remove the "orphaned" suffix. This could be addressed in another issue, since again it does not affect correctness, but rather only performance in case of a large trie with many deleted terms (that without the proposed refactoring stay in the trie, even though they point to empty document sets).

I am confident that @micheleriva understands what I am pointing at, but otherwise I can offer some more explanation.

packages/lyra/src/prefix-tree/trie.ts
```diff
@@ -112,12 +112,12 @@ export class Trie {
     if (!word) return false;
 
     function removeWord(node: TrieNode, _word: string, docID: string): boolean {
-      const [nodeWord /**_docs*/] = node.getWord();
+      const [nodeWord, docIDs] = node.getWord();
 
       if (node.end || (exact && node.end && nodeWord === word)) {
```
lucaong, line 117: not related to this PR (rather to #41), but still worth noting: this condition is completely equivalent to just `node.end`: that's because `x || (x && y)` is the same as just `x`. In the specific case, if `node.end` is true, the condition short circuits before the `||`. If it is false, then the part after the `||` is also false, because it's a chain of `&&` including `node.end`. Therefore, the whole `exact` logic can be removed with no effect.

Whether the resulting condition after removing the part after `||` is correct, I am not 100% sure (I would advise having more extensive tests), but for sure it is equivalent to the current one.

lucaong (commented):
This fixes the issue, but for the "wrong reason" :) See the comment.

PR #46, opened 2022-07-24
perf: prevent potential overhead on falsy value

thomscoder:
This should prevent unnecessary linear time computations 
(the potential perf improvement is proportional to the docIDs size)


micheleriva:
Thank you @thomscoder 🙏

PR #57, opened 2022-07-29
fixed example into README file for case sensitive search

AlessandroMinoccheri:
Hi all, 
I tried for the first time this library and following the README example, I saw that searching with `if` word returns nothing.
Searching with `If`returns expected results because the search is case-sensitive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ After the data has been inserted, you can finally start to query the database.
 
 ```js
 const searchResult = search(db, {
-  term: 'if',
+  term: 'If',
   properties: '*'
 });
 ```


README.md
```diff
@@ -61,7 +61,7 @@ After the data has been inserted, you can finally start to query the database.
 
 ```js
 const searchResult = search(db, {
-  term: 'if',
+  term: 'If',
```
thomscoder, line 64: Hi @AlessandroMinoccheri 👋🏽 
I've looked into it (as I'm writing the docs for this). 
It seems Lyra starts being case insensitive by default only when the `term`.length > 2.

```js
const db = create({
  schema: {
    title: 'string',
  }
});

insert(db, {
  title: 'What If',
});

insert(db, {
  title: 'what if',
});

const searchResult = search(db, {
  term: 'if',
  properties: '*'
});
```
Result

```sh
{
  elapsed: 87334n,
  hits: [ { id: '23490458-8', word: 'what if' } ],
  count: 1
}
```

When the term.length > 2

```js
const db = create({
  schema: {
    word: 'string',
  }
});

insert(db, {
  word: 'air',
});

insert(db, {
  word: 'Air',
});

const searchResult = search(db, {
  term: 'air',
  properties: '*'
});
```
result

```bash
{
  elapsed: 89416n,
  hits: [
    { id: '23309215-1', word: 'air' },
    { id: '23309215-5', word: 'Air' }
  ],
  count: 2
}
```

Interesting.

AlessandroMinoccheri: @thomscoder you are right! It could be interesting to fix this problem. I am trying to understand the logic of the search.

AlessandroMinoccheri: I opened an issue for this problem in the meanwhile

thomscoder: > I opened an issue for this problem in the meanwhile

Perfect, thank you ! 😄

AlessandroMinoccheri: @thomscoder I tried to fix it but I don't know if is the right way, if not I would like to discover together the right way.
https://github.com/nearform/lyra/pull/59

micheleriva:
Question is... do we want case sensitive matches at all?

AlessandroMinoccheri:
With the new release, this problem is fixed.

Issue #58, opened 2022-07-29
Search case-sensitive for terms with a length less than 3


AlessandroMinoccheri:
**Describe the bug**
It seems that the search is case-sensitive for words with a length < 3

**To Reproduce**
Insert records like: 
* ```if you are interested```
* ```if you will go out..```
* search for the term `If`
* you retrieve 0 result
* if you search for the terms: `if you`
* you get results

**Expected behavior**
I am expecting that the search is case-sensitive or not with a specific rule

related PR: #57 (but it doesn't fix the problem)

micheleriva:
@AlessandroMinoccheri my question is... do we want case-sensitive searches at all?

AlessandroMinoccheri:
This is the main question @micheleriva 
We can have a default behavior and a parameter to specify if the search is case-sensitive or not IMHO.
But we need to have a default and consistent behavior in the code. 
This PR could be useful to understand it not only to solve the problem.

AlessandroMinoccheri:
If I understood well the biggest problem is that with two characters we have a behavior with more another behavior.

micheleriva:
Adding support for case-sensitive matches may lead us to extra effort with minimal benefits... with the latest PR I merged, we are converting all the tokens to lowercase so the problem should be resolved (in fact, I had to edit a couple of tests that were wrong for the reason you're suggesting).

If we can find a way to add case-sensitive matches with minimum efforts, I'm ok with that!

AlessandroMinoccheri:
I tried the behavior and it seems to work fine.
Can we explicitly write that the search is case-insensitive? Or is already written?

micheleriva:
I was waiting for #24 to be completed before updating the overall documentation!

AlessandroMinoccheri:
Ok, so I can close this issue!
Thank you :)

PR #59, opened 2022-07-29
fixed case sensitive search

AlessandroMinoccheri:
In this pull request, I try to fix issue #58 checking for lowercase and uppercase character


micheleriva:
Thanks for your contribution @AlessandroMinoccheri!

I think we may want to add the possibility to select whether a search should be case sensitive or not, what do you think?

Something like:

```js
search(db, {
  term: 'Hello',
  caseSensitive: true
})
```

I don't particularly like the `caseSensitive` property name, so I'm open to suggestions 🙂

Also, may I ask you to add more tests?

AlessandroMinoccheri:
`caseSensitive` is good enough for me, I think it's better to close this PR and open a specific issue and a PR for this new behavior, what do you think @micheleriva ?

micheleriva:
Let's do that, thank you!

Issue #29, opened 2022-07-18
Search method returns undefined elements


mateonunez:
**Describe the bug**
Searching by a term that doesn't exist, Lyra returns `X` undefined elements.

The error also occurs using `limit`, `property` or `exact` params.

**To Reproduce**
1. `lyra.search({ term: "whatever" })`

**Expected behavior**
Lyra shouldn't return `X` undefined elements. Instead of that should return an empty array.

**Screenshots**

![image](https://user-images.githubusercontent.com/11861080/179612379-4981aff4-2baa-4101-84e7-d1946299a290.png)

**Desktop:**

- OS: Windows 11
- Environment: Node
- Node: v18.6.0
- pnpm: v7.5.1
- Lyra: v0.0.1-beta-12

micheleriva:
When retrieving data, we create an array of `x` length, where `x` is the `limit` search param. That way we avoid unnecessary dynamic allocations when adding data to the `results` array, but of course is causing this problem

```js
    const results: RetrievedDoc<TSchema>[] = Array.from({
      length: limit,
    });
``` 

Thanks for opening the issue!

mateonunez:
And what would be the correct way to fix it? Create an Array that also allows dynamic allocations? Or filter the Array before the return?

micheleriva:
I'd say... let's benchmark these operations with different array sizes. Let's see when is convenient filtering vs dynamic arrays given the docs we have

PR #73, opened 2022-08-01
Add Lyra implementation with Vue

mateonunez:
This PR would add an example of Lyra usage with Vue framework.

The version of Vue used is 3. TypeScript has not been implemented in this project as it is not yet optimized like React (this can be discussed).

The example is the same made for `React` and uses the same libraries: `unocss` and `vite`


packages/examples/with-vue/src/App.vue
```diff
@@ -0,0 +1,113 @@
...
+          properties: "*",
+        });
+
+        this.elapsedTime = formatNanoseconds(getNanosecondsTime() - timeStart);
+
+        const pokemons = result.hits.map(({ id: hitID }) =>
+          this.pokemon?.find(({ id: pokemonID }) => pokemonID === hitID),
+        );
+
+        if (pokemons.length && !pokemons.some(x => !x)) {
```
thomscoder, line 62:
```js
!pokemons.some(x => !x);
```
This got me wondering if Lyra can return `null` or `undefined` values
```js
const pokemons = [null];
```
for example 🤔

thomscoder (commented):
LGTM! 💯

packages/examples/with-vue/src/App.vue
mateonunez: Yep, Lyra sometimes returned undefined values (#29). That issue was fixed, so we could consider this control unnecessary but it would be better to wait for the stable version before removing it.

thomscoder: Perfect!
Nice job!

micheleriva:
thanks! lgtm

PR #81, opened 2022-08-03
Demo page improvements and restyling

paolo-cargnin:
So I read the tweet of Michele, and I tried to improve the demo page as I could without having any specification about the needs.

I aligned better the form elements and changed the header of the page (imho this style is cleaner and more modern, maybe we could adopt it even in the homepage?). 
I also added autofocus on the search input. 
Of course if it is not as you guys were thinking it, I can improve it.

The main part now that should be improved is the search results, but about this... Can I suggest to use a better structured database? The categories in this example are not easy to display and organize since they are not always the same. Maybe using some more "mainstream" data could be nice   (I was thinking movie quotes) ? Dunno.

Let me know!


ShogunPanda:
Hi Paolo! (nice name XD) 
In #82 I also redesigned and improved the demo page. Do you mind reapplying your changes after the current main?

paolo-cargnin:
Yeah of course! I'll do it this evening ;)

paolo-cargnin:
Just to be sure, I added this heading to the demo page:
image


Do you prefer this style or the one that there is right now?

ShogunPanda:
@paolo-cargnin What about having that gradient in lieu of the current pink background in the hero and use regular text?

paolo-cargnin:
Ok done. Most of the usability features were already done by you. I list here the super small things that I have done:
- added a progress bar for the indexing
- heading as described by Paolo
- Fixed a "category 2" entry on the events.ts file (so it will render "&" instead of &amp)
- Aligned a bit better the result block
Let me know!

packages/docs/src/components/LyraDemo/style.module.css
```diff
@@ -53,9 +58,43 @@ select {
   flex-direction: column;
   background-color: var(--ifm-color-primary);
   color: white;
-  margin-bottom: 2rem;
   padding: 2rem;
   text-align: center;
+  position:relative;
+}
+.hero > *{
+  position:realative;
```
ShogunPanda, line 66: Typo. Also, can we avoid the star selector?

ShogunPanda (approved):
Just a minor thing. Rest LGTM, thanks!

micheleriva (approved):

PR #86, opened 2022-08-05
fix: intelliSense not working for some params of insert/search

iShibi:
The `ResolveSchema<S>` and `SearchProperties<S>` types for insert's `doc` and search's `params.properties` parameters respectively are defined in the following `types.d.ts` file:

https://github.com/nearform/lyra/blob/2a46bf5f024e90212b5d528fb2162b659392c811/packages/lyra/src/types.d.ts#L14

https://github.com/nearform/lyra/blob/2a46bf5f024e90212b5d528fb2162b659392c811/packages/lyra/src/types.d.ts#L18

The build generated by tsc does not include this declaration file, resulting in no autocompletion/suggestion for the above parameters. This PR converts this declaration file into a normal `.ts` file and lets the compiler generate the corresponding `.d.ts` file for it in the build.


micheleriva:
thank you!

micheleriva (approved):

PR #87, opened 2022-08-05
chore: do not publish source files to npm

iShibi:
`30.3 MB` Unpacked Size 😨😅

This change will make sure that only the `dist` folder and other files like `LICENSE`, `README`, and `package.json` get published to npm.


micheleriva:
You're right, but we had some problems in the past while publishing the `dist` folder only. Let me try pull your branch and test a publication on Verdaccio 🙂

micheleriva:
@iShibi for some reason this is not working. Would you mind testing a publication against a local version of Verdaccio?

iShibi:
Publishing works fine for me after bumping the version to `0.0.4` since `<=0.0.3` are already available on npm:

lyra-1
--- snip ---
lyra-2

micheleriva:
@iShibi I was able to publish on Verdaccio too, but after trying to install from there, I don't get the `dist` folder

iShibi:
That's weird, because everything works fine for me:

test

micheleriva:
Thanks for the heads up, I'll verify this with my team

iShibi:
@micheleriva Any updates on this?

I tested verdaccio a little more and found out that without the `files` field it does not publish gitignored folders/files, which in this case are `coverage`, `node_modules`, and <ins>`dist`</ins>. Just wanted to double check whether you had this `files` field in the `package.json` while publishing.

simonireilly:
It could be a valuable part of this to run `pnpm run pack` as part of the CI pipeline.

I am not suggesting continuous deployment/release, but just ensuring the contents of the package are in a releasable state. 

This will give a debug level log of the assets that will pack and we can trust CI instead of reasoning around the differences on specific setups :+1:

micheleriva:
Fixed in v0.1.3

PR #91, opened 2022-08-08
docs: add inline documentation

iShibi:
I have added docs only for the four major functions. Once the writing style gets a green flag I would be happy to document all the other stuff.

### DIFF
--- a/packages/lyra/src/lyra.ts
+++ b/packages/lyra/src/lyra.ts
@@ -174,6 +174,18 @@ function getDocumentIDsFromSearch<S extends PropertiesSchema>(
   return Array.from(ids);
 }
 
+/**
+ * Creates a new database.
+ * @param properties Options to initialize the database with.
+ * @example
+ * // Create a database that stores documents containing author and quote fields.
+ * const db = create({
+ *   schema: {
+ *     author: 'string',
+ *     quote: 'string'
+ *   }
+ * });
+ */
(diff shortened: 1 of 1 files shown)


RafaelGSS (commented):
I believe the inline documentation is useful just for the methods exported by this library. Internal methods should have their naming clear enough

iShibi:
By other stuff I meant the `interfaces`. Doing so would allow the editor to show what a particular field in an object parameter does.

RafaelGSS (commented):
Including `SearchResult` would be great, yeah.

RafaelGSS (commented):
Most of the inline documentation is pretty redundant to me. Example:
```ts
export interface Lyra<S extends PropertiesSchema> extends Data<S> {
  /**
   * The default language analyzer in use.
   */
  defaultLanguage: Language;
  /**
   * The structure of documents stored in the database.
   */
  schema: S;
  edge: boolean;
}
```

You don't need to add inline documentation to everything, just mention unclear fields. For instance: 

```ts
/**
   * The maximum [levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance)
   * between the term and the searchable property.
   */
  tolerance?: number;
```

is a great addition. As said previously, I'd keep just the inline doc for exported functions + unclear fields (`tolerance`)

RafaelGSS (approved):

PR #95, opened 2022-08-15
refactor: fix some minor linter warnings

castarco:
**EDIT:** Now I'm not so sure if it's a good idea to propose these changes, as `repl.ts` seems to be kinda "out of reach" for the type checker, and I might be introducing some difficult-to-detect problems until that is solved in the first place.

The changes of this PR are merely cosmetic, but I find it a good thing to have a smaller amount of warnings when we run the linter :) .

### DIFF
--- a/packages/development-repl/repl.ts
+++ b/packages/development-repl/repl.ts
@@ -1,7 +1,7 @@
 import * as fs from "node:fs";
 import * as readline from "node:readline";
 import * as util from "node:util";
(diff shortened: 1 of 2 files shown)


micheleriva (approved):

micheleriva:
thank you 🙏

castarco:
Ups, I should have marked it as draft, i pushed an extra refinement just now 😓

PR #102, opened 2022-08-24
fix: check for hrtime func availability

### LINKED ISSUE #101: Process.hrtime is undefined with Next.js since 0.0.5
**Describe the bug**

On upgrade to `0.1.1` (from `0.0.4`) the search function spits out a process error in Next.js

```
TypeError: process.hrtime is undefined
```

Guessing this might be due to the changes in  https://github.com/LyraSearch/lyra/commit/b51456c2c22b99337f492942697ad32299701f9b which were added to `0.0.5`

Here's a sandbox reproduction using the most popular TS/Next.js configuration

https://codesandbox.io/s/magical-carlos-3p11ow

Sorry I can't be more helpful and submit a PR, but I'm mostly a designer! Great little lib, ty for putting it together.

castarco:
Fixes #101

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -35,7 +35,7 @@ export function formatNanoseconds(value: number | bigint): string {
 }
 
 export function getNanosecondsTime(): bigint {
-  if (typeof process !== "undefined") {
+  if (typeof process !== "undefined" && process.hrtime !== undefined) {
     return process.hrtime.bigint();
   }
 


micheleriva (approved):
lgtm

micheleriva:
Thank you @castarco !

src/utils.ts
```diff
@@ -35,7 +35,7 @@ export function formatNanoseconds(value: number | bigint): string {
 }
 
 export function getNanosecondsTime(): bigint {
-  if (typeof process !== "undefined") {
+  if (typeof process !== "undefined" && process.hrtime !== undefined) {
```
codyzu, line 38: it feels a little inconsistent here... do we need an issue to use `!== undefined` everywhere? Using `typeof x !== "string"` feels kind of old school, right???

castarco: The first part of the conditional makes sense because, if `process` is not declared (as opposed to be `undefined`, then comparing it against undefined, or trying to do something like `if (process)` raises an error.

It is true, though, that the second part of the conditional could have been simplified.

PR #106, opened 2022-08-27
feat(lyra): adds batchInsert async method

micheleriva:
This PR adds `batchInsert`, an async method capable of inserting a large number of documents without blocking the
event loop.

Adapted from the code written by @ShogunPanda as part of the benchmark suite (all credits to him): https://github.com/LyraSearch/lyra/commit/d0f95d789ed80de2f8ebc31eedee5e72c0a27877

### DIFF
--- a/src/lyra.ts
+++ b/src/lyra.ts
@@ -47,6 +47,10 @@ export type InsertConfig = {
   language: Language;
 };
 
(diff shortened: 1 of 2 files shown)


ShogunPanda (commented):
LGTM!

PR #109, opened 2022-08-28
feat(stemmer): adds english stemmer

micheleriva:
This PR aims to add an English stemmer to Lyra.
The API is designed in such a way that if a stemmer is not available for a selected language, the stemming process will be skipped.

The stemming algorithm is based on the following paper: http://www.tartarus.org/~martin/PorterStemmer.

### DIFF
--- a/src/tokenizer/stemmer/en.ts
+++ b/src/tokenizer/stemmer/en.ts
@@ -1,0 +1,189 @@
+const step2List = {
+  ational: "ate",
+  tional: "tion",
+  enci: "ence",
+  anci: "ance",
+  izer: "ize",
+  bli: "ble",
+  alli: "al",
+  entli: "ent",
+  eli: "e",
+  ousli: "ous",
+  ization: "ize",
+  ation: "ate",
+  ator: "ate",
+  alism: "al",
+  iveness: "ive",
+  fulness: "ful",
+  ousness: "ous",
+  aliti: "al",
+  iviti: "ive",
+  biliti: "ble",
+  logi: "log",
+};
+
+const step3List = {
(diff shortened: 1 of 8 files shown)


src/tokenizer/stemmer/en.ts
```diff
@@ -0,0 +1,189 @@
+const step2List = {
```
ShogunPanda, line 1: Non blocking comment on this file.

If you can, it would be great to extract all constant RegExp expression at the module level and declare them as RegExp when possible so we might get a small to tiny performance improvement.

ShogunPanda (approved):
LGTM!

src/lyra.ts
```diff
@@ -18,6 +20,20 @@ export type PropertiesSchema = {
...
 };
 
+export type TokenizerConfig = {
+  enableStemming?: boolean;
+  enableStopWords?: boolean;
+  customStopWords?: ((stopWords: string[]) => string[]) | string[];
+  stemmingFn?: (word: string) => string;
+};
+
+export type TokenizerConfigExec = {
```
ShogunPanda, line 30: You can streamline this as `export type TokenizerConfigExec = Required<TokenizerConfig>`

ShogunPanda (approved):
Few minor things.

src/lyra.ts
micheleriva: Nice catch!

micheleriva: Oh, one problem is that `stemmingFn` can be undefined (to fallback to the default stemming function). `Required` is not compatible with `undefined` values

PR #110, opened 2022-08-28
feat: implement hook system

RafaelGSS:
This PR implements the foundation for the Hooks System.

The main changes are:
- `create` function accept and validate `hooks`
- a new function `insertWithHooks` is exported
- `insertBatch` uses `insertWithHooks`, hooks is now available


src/lyra.ts
```diff
@@ -101,6 +113,31 @@ export type RetrievedDoc<S extends PropertiesSchema> = ResolveSchema<S> & {
...
 
+function validateHooks(hooks?: Hooks): void | never {
+  if (hooks) {
+    if (typeof hooks !== "object") {
+      throw new Error("Invalid hooks object");
+    }
+
+    const invalidHooks = Object.keys(hooks).filter(hook => !SUPPORTED_HOOKS.includes(hook));
+    if (invalidHooks.length) {
+      throw new Error(`The following hooks aren't supported. Hooks: ${invalidHooks}`);
```
micheleriva, line 124: I'd move all the error messages into the `src/errors.ts` file

src/lyra.ts
```diff
@@ -326,7 +407,7 @@ export async function insertBatch<S extends PropertiesSchema>(
 
       for (const line of batch) {
         try {
-          insert(lyra, line, config);
+          await insertWithHooks(lyra, line, config);
```
micheleriva, line 410: As you can see, `insertBatch` returns a `Promise<void>`. Maybe we want to return a list of IDs or something? What would be the benefit of using hooks here otherwise? wdyt?

RafaelGSS: The advantage is that you can control when and what to do after each insertion, for example. But, since the `insertBatch` signature didn't change, that's sounds a reasonable change.

I'd not return the list of IDs, since we have the hooks they can use it if they want, otherwise, we'll need to increase the memory consumption to do that.

micheleriva (approved):
lgtm

src/lyra.ts
ravaan: Hey @micheleriva @RafaelGSS  can we return a list of ids from insertBatch. There is no option for a custom id, so insertBatch is not of use if you want to maintain a mapping with your database for invalidation of data.

ravaan: If all the create/update functions return id's then insertBatch  should also return id's preserving order

ravaan (commented):
Return array of Id's from `insertBatch` preserving order

src/lyra.ts
ravaan: The memory argument is valid, maybe if not add return of id's here then add something in the documentation pointing towards `insertWithHooks` function when you need to get list of id

ShogunPanda: This will be resolved when we rehaul the hook system in the future with more hooks. For now I agree not to increase memory consumption.

ravaan: @ShogunPanda I am not sure if the new hooks system has been added or not, but what do you think about my point of adding something in the documentation pointing towards `insertWithHooks` when you need to get a list of id? 

I was stuck again today on the same thing with my old code and was reminded of this!

ShogunPanda: @ravaan insertWithHooks does not exist anymore. Nowadays you can provide `afterInsert 
` when doing the `create` to achieve the same result.

See: https://docs.oramasearch.com/internals/components#afterinsert

Does it answer your question?

PR #113, opened 2022-08-31
fix: replace setImmediate with setTimeout

sidwebworks:
Currently, the `batchInsert` function uses `setImmediate` to schedule the recursive `insert` calls to avoid blocking the event loop.

However, this causes the API to break in the browsers where `setImmediate` is not defined as it's a Node API.

This PR fixes that issue by using `setTimeout` with a delay of `0`.

cc// @micheleriva


micheleriva (approved):
Thanks! Lgtm

PR #120, opened 2022-09-05
feat(package): add proper support for webpack 4 which is used by expo

codyzu:
* Expo uses webpack 4 for web builds. They have a draft PR to upgrade to webpack 5: https://github.com/expo/expo-cli/pull/3763
* webpack 4 uses Acorn 6.x for parsing, which does not support the latest ECMA script features (such as optional chaining). webpack decided to integrate Acorn >7.3.0 only into webpack 5: https://github.com/webpack/webpack/issues/10227
* The above 2 points mean that until expo updates to webpack 5, expo web builds does not support language features such as optional chaining `?.` that are used by Lyra.
* Solution A (this PR): fix Lyra by adding a new typescript build for "browser" (which is used by [webpack 4](https://v4.webpack.js.org/configuration/resolve/#resolvemainfields) _and_ [webpack 5](https://webpack.js.org/configuration/resolve/#resolvemainfields)) that targets `ES2019` (the latest compatible with Acorn 6.x). It will also be used by (non expo) react native projects via the [metro loader](https://vitejs.dev/config/shared-options.html#resolve-mainfields). Since next.js uses webpack 5, it will likely pick up the "browser" entry for web builds too. [Vite](https://vitejs.dev/config/shared-options.html#resolve-mainfields) uses the "module" entry and won't be affected. Same for node.
* Solution B (not implemented): Change the existing cjs and esm Lyra builds to target `ES2019`. I think we prefer "pure", non transpiled, ECMAScript when possible.
* Solution C (bad devex): Force application builders to [transpile Lyra manually](https://github.com/expo/expo/issues/7036#issuecomment-634017272) when building with expo. I don't think this is reasonable because expo is a "batteries included" RN development framework.
* once https://github.com/expo/expo-cli/pull/3763 is merged, we can undo this special build
* I added `concurrently` to build the now 3 builds in parallel


package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Fast, in-memory, typo-tolerant, full-text search engine written in TypeScript",
   "main": "./dist/cjs/lyra.js",
   "module": "./dist/esm/lyra.js",
+  "browser": "./dist/rn/lyra.js",
```
micheleriva, line 7: I'd make the output directory name easier to read so that people could import it as:

```js
import { create } from '@lyrasearch/lyra/react-native'
```

tsconfig.webpack.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "./tsconfig.json",
+  "include": [
+    "src"
+  ],
+  "compilerOptions": {
+    "target": "ES2019",
```
micheleriva, line 7: Isn't the optional chaining introduced with ES2019? Maybe we want to compile to ES6 for React Native

package.json
```diff
@@ -22,7 +23,8 @@
     "build:lyra": "cd packages/lyra && pnpm build",
     "build:cjs": "tsc --project tsconfig.cjs.json",
     "build:module": "tsc --project tsconfig.esm.json",
-    "build": "rimraf ./dist && pnpm build:module && pnpm build:cjs",
+    "build:webpack": "tsc --project tsconfig.webpack.json",
```
micheleriva, line 26: Are we planning to use `webpack` for something different than React Native (i.e., browser bundles)? Otherwise I'd change this to `build:react-native`

micheleriva:
Thanks for the PR @codyzu !

I'd go with solution A, with a specific build for react-native. Hopefully, we can get rid of it once it supports webpack 5

package.json
codyzu: No really not, but I think indirectly it will be picked up by anyone using next.js (I can't find docs on this) or create-react-app (uses webpack 5)... that's why I was trying to keep it generic. maybe renaming it to `browser` everywhere would be more intuitive??? What do you think?

micheleriva: I tested this in Next.js and it works well (Next.js uses Webpack 5)!

In any case, we can go with browser and make it more generic 👍

tsconfig.webpack.json
codyzu: Nope, in ES2020. There are resources [here](https://kangax.github.io/compat-table/es2016plus/) and [here](https://github.com/sudheerj/ECMAScript-features).

And I got the idea here: https://stackoverflow.com/a/66422985

More importantly, I tested it with my expo app and it works with `ES2019` 😉

codyzu:
We can create an open issue the tracks the PR in expo... that way we don't forget. I'll do that now.

micheleriva:
thanks! lgtm

PR #125, opened 2022-09-06
build(package): use npm for build to allow for vanilla install from git and github

codyzu:
If installing via git or github (not the published package on npm), it doesn't make sense to require pnpm. For better or worse, npm is the standard.


micheleriva (approved):
thanks!

Issue #126, opened 2022-09-06
Error: Cannot find module '../stemmer/cjs/en'


mateonunez:
**Describe the bug**

The new release can't find module `'../stemmer/cjs/en'`, imported in `lyra.js:6`

**To Reproduce**
1. npm init --yes
2. npm install @lyrasearch/lyra@0.2.2
3. npm run start

**Expected behavior**

When Lyra is build, the stemmer path becomes wrong: it adds one level on the dist folder.

![image](https://user-images.githubusercontent.com/11861080/188730104-ec909da4-de43-4983-aa08-27c27f18bd08.png)

**esm stemmer import**
![image](https://user-images.githubusercontent.com/11861080/188730340-8e3b7647-134d-4e5d-be7f-479be3e082eb.png)

_the import should be: `import { stemmer } from "../../stemmer/cjs/en"`_

**cjs stemmer import**
![image](https://user-images.githubusercontent.com/11861080/188730400-cf9d65c6-fd19-42df-bab3-5b2d29a8a2e8.png)

_the import should be `const en_1 = require("../../stemmer/cjs/en");`_

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Node: v18.7.0
 - Lyra: v0.2.2

AliMejbar:
Same Problem.
OS: macOS 13.0
Node: v18.7.0
Lyra: v0.2.2

micheleriva:
Thanks for opening this, would you be available for opening a PR? 🙏

PR #127, opened 2022-09-07
fix(stemmer): fix stemmer installation path

micheleriva:
hotfix for #126

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@lyrasearch/lyra",
-  "version": "0.2.2",
+  "version": "0.2.3",
   "description": "Fast, in-memory, typo-tolerant, full-text search engine written in TypeScript",
   "main": "./dist/cjs/lyra.js",
   "module": "./dist/esm/lyra.js",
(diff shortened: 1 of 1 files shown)


micheleriva:
@codyzu any chance you can fix this? 😔

package.json
```diff
@@ -19,10 +19,10 @@
     "format": "prettier --write \"**/*.{ts,js}\"",
     "lint": "pnpm run format && eslint . --ext .js,.ts",
     "prepare": "husky install && npm run build",
-    "build:cjs": "tsc --project tsconfig.cjs.json",
-    "build:module": "tsc --project tsconfig.esm.json",
+    "build:cjs": "tsc --project tsconfig.cjs.json && cp -R stemmer/cjs/en.js dist/stemmer/cjs/en.js && cp -R stemmer/cjs/en.d.ts dist/stemmer/cjs/en.d.ts",
+    "build:module": "tsc --project tsconfig.esm.json && cp -R stemmer/esm/en.js dist/stemmer/esm/en.js && cp -R stemmer/esm/en.d.ts dist/stemmer/esm/en.d.ts",
     "build:browser": "tsc --project tsconfig.browser.json",
-    "build": "rimraf ./dist && concurrently 'npm:build:module' 'npm:build:cjs' 'npm:build:browser'",
+    "build": "rimraf ./dist && mkdir -p dist/stemmer/{cjs,esm} && concurrently 'npm:build:module' 'npm:build:cjs' 'npm:build:browser'",
```
mateonunez, line 25: Replace that main build command with:

```js
"build": "rimraf ./dist && concurrently 'npm:build:module' 'npm:build:cjs' 'npm:build:browser' && cp -R stemmer/ dist/stemmer",
```

The `dist` path will contain _browser_, _cjs_, _esm_ and _stemmer_.

Is not necessary change `build:cjs` and build:esm` commands.

codyzu: this probably would have fixed the issue, the downside is that `cp` is not windows compatible. Probably better to use something like https://www.npmjs.com/package/copyfiles

In any case, I think a better solution is to let the typescript compiler handle this. See #128

codyzu:
> @codyzu any chance you can fix this? 😔

@micheleriva I took a different approach in #128. With the new approach, you can _only_ include the esm build from snowball 🎉 

I recommend closing this in favor of #128 when you have a chance to look

micheleriva:
Thanks Cody you're a hero

PR #128, opened 2022-09-07
fix(stemmer): fix handling of stemmer imports

codyzu:
hotfix for #126

* enables the [TS `allowJS`](https://www.typescriptlang.org/tsconfig#allowJs) flag so that the TS compiler copies the js files from stemmer as suggested [here](https://vccolombo.github.io/blog/tsc-how-to-copy-non-typescript-files-when-building/)
* added a "smoke" test that runs a basic search after importing lyra from `dist`. This is an attempt to avoid errors like this in the future. I added this test to the CI also.
* removes the cjs build of the stemmer and renames the esm build to `lib`. Now that the typescript compiler handles importing, it should only be esm

**Notes**
1. The npm run script `ci` with coverage checks isn't actually run by the CI and notably the coverage checks are failing. Probably want to fix this in the future


codyzu:
replaces and closes #127

codyzu:
fixes #126

ShivamJoker:
I am facing the same problem can this be merged soon?

micheleriva:
We'll get this fixed by this evening

codyzu:
If you're blocked and don't want to roll back to the previous release or wait until tonight for a new release, you can try out my fix by removing Lyra from your `package.json` and then `npm install "codyzu/lyra#fix-stemmer-inclusion"`

micheleriva:
Thank you Cody, I'm gonna merge this

micheleriva:
@ShivamJoker this is fixed

ShivamJoker:
Awesome I'll upgrade and let you know

PR #130, opened 2022-09-10
chore: release automation

simoneb:
Adds https://github.com/nearform/optic-release-automation-action and takes the chance to refactor the existing CI script, which was unnecessarily complex.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,38 @@
+name: release
+
+on:
+  workflow_dispatch:
+    inputs:
+      semver:
+        description: "The semver to use"
+        required: true
+        default: "patch"
+        type: choice
+        options:
+          - patch
+          - minor
+          - major
+      tag:
+        description: "The npm tag"
+        required: false
+        default: "latest"
+  pull_request:
+    types: [closed]
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: pnpm/action-setup@v2
+        with:
+          version: 7
+      - uses: nearform/optic-release-automation-action@v3
+        with:
+          npm-token: ${{ secrets[format('NPM_TOKEN_{0}', github.actor)] || secrets.NPM_TOKEN }}
+          optic-token: ${{ secrets[format('OPTIC_TOKEN_{0}', github.actor)] || secrets.OPTIC_TOKEN }}
+          semver: ${{ github.event.inputs.semver }}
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -1,37 +1,22 @@
...
-    - name: Install dependencies and run tests
-      run: pnpm install && pnpm run test && pnpm run test:smoke
+      - uses: actions/checkout@v3
+      - uses: pnpm/action-setup@v2
+        with:
+          version: 7
+      - uses: actions/setup-node@v3
+        with:
+          node-version-file: .nvmrc
+          cache: pnpm
```
simoneb, line 18: the simplification overall boils down to this. setup-node natively supports caching pnpm, so there's no need for a custom setup

.github/workflows/tests.yml
```diff
@@ -1,37 +1,22 @@
 name: Tests
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
simoneb, line 6: this is the recommended setup to avoid running this workflow too many times.

.github/workflows/release.yml
```diff
@@ -0,0 +1,38 @@
...
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: pnpm/action-setup@v2
+          with:
+            version: 7
+      - uses: nearform/optic-release-automation-action@v3
+        with:
+          npm-token: ${{ secrets[format('NPM_TOKEN_{0}', github.actor)] || secrets.NPM_TOKEN }}
+          optic-token: ${{ secrets[format('OPTIC_TOKEN_{0}', github.actor)] || secrets.OPTIC_TOKEN }}
```
simoneb, line 32: these secrets need to set up in the repo and a decision needs to be made about who can release.

micheleriva: The `OPTIC_TOKEN` is not required if I use an `npm` publish token right?

simoneb: correct. the npm automation token allows you to bypass 2FA. if you want to go that way the release process is simpler (though less secure of course). you would still need to decide whether to have a single NPM token or one per person who can release. the second option allows to restrict who can release to only the people who have a NPM_TOKEN_{username} configured in the repo secrets

micheleriva: Perfect, I already set up an `NPM_TOKEN` for the next release. Next week we'll catch up to decide who has the rights to make a release, thanks a lot for this PR

PR #131, opened 2022-09-11
feat(lyra): compute levenshtein metric within a given tolerance

jkomyno:
As discussed privately, I have noticed that the `levenshtein` distance is currently only used as a comparison metric to check whether two strings have edit distance less than a given `tolerance` threshold.

Computing the `levenshtein` metric in the general case takes time `O(max(m, n))` and space `O(max(m, n))`, where `m` and `n` is the length of the two input strings. When we're only interested in retrieving the exact edit distance when it's below or equal to a given `tolerance` threshold, more time-efficient methods exist.

Given

```ts
let term: string;
let word: string;
let tolerance: number; // small, non-negative integer
```

this PR turns the comparison

```ts
if (levenshtein(term, word) <= tolerance) {
  // ...
}
```

into the following:

```ts
if (boundedLevenshtein(term, word, tolerance).isBounded) {
  // ...
}
```

The core algorithm this PR takes inspiration from has already been developed in the [`talisman`](https://github.com/Yomguithereal/talisman/blob/86ae55cbd040ff021d05e282e0e6c71f2dde21f8/src/metrics/levenshtein.js#L218-L340) project, which is MIT licensed.

It'd be interested to see how the changes introduced by this PR will influence the benchmarks. I expect a marginal performance increase as the size of the inputs grows.

## Alternative not considered in this PR

As the size of the input scales, computing an exact comparison metric may not be needed. Luckily, the literature offers a plethora of polylogarithmic approximation algorithms to choose from. Should you be interested in this slightly less precise alternative, I'd suggest to take a look at [Andoni *et al.*](https://onak.pl/papers/focs_2010-approximating_edit_distance.html)'s method.

### DIFF
--- a/src/levenshtein.ts
+++ b/src/levenshtein.ts
@@ -1,3 +1,141 @@
+type BoundedMetric = {
+  isBounded: boolean;
+  distance: number;
+}
+
+/**
+ * Computes the Levenshtein distance between two strings (a, b), returning early with -1 if the distance
+ * is greater than the given tolerance.
+ * It assumes that:
+ * - tolerance >= ||a| - |b|| >= 0
+ */
+export function boundedLevenshtein(a: string, b: string, tolerance: number): BoundedMetric {
+  const distance = _boundedLevenshtein(a, b, tolerance);
+  return {
+    distance,
+    isBounded: distance >= 0,
+  };
+}
+
+/**
+ * Inspired by:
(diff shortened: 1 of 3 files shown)


micheleriva:
Hi @jkomyno,
Thank you so much for this PR, this is brilliant!

About the benchmarks, I forgot we removed the `benchmark` script from this repo, so I'll run them manually and update you on the results

micheleriva (approved):

Issue #136, opened 2022-09-17
Browser build doesn't export methods


mateonunez:
**Describe the bug**
The current version of Lyra, ran in a browser doesn't work.

```bash
Uncaught SyntaxError: The requested module './src/lyra.js' does not provide an export named 'create' (at index.html:5:16)
```

I got the same error with `search` and `insert` methods.

**To Reproduce**
1. Create a new index file importing a browser release: [https://unpkg.com/@lyrasearch/lyra@0.2.4/dist/browser/src/lyra.js](https://unpkg.com/@lyrasearch/lyra@0.2.4/dist/browser/src/lyra.js)
2. Paste the following code:
```html
<!DOCTYPE html>
<html>
  <body>
    <script type="module">
      import { create, search, insert } from "https://unpkg.com/@lyrasearch/lyra@0.2.4/dist/browser/src/lyra.js";

      const db = create({
        schema: {
          author: "string",
          quote: "string",
        },
      });

      insert(db, {
        quote: "It is during our darkest moments that we must focus to see the light.",
        author: "Aristotle",
      });

      const searchResult = search(db, {
        term: "if",
        properties: "*",
      });

      console.log(searchResult)
    </script>
  </body>
</html>
```
3. Open the file

**Expected behavior**
Lyra should be used in any runtime

**Desktop (please complete the following information):**
 - Lyra: **v0.2.4**

**Additional context**
_EDITED:_ A good `tsc` alternative to building the browser release is to use `esbuild` (used by Lyra a few versions ago), I've created a [branch](https://github.com/mateonunez/lyra/tree/perf/browser-build) with a working browser release.

micheleriva:
Thank for opening this @mateonunez!

@codyzu, looking at the bundle at `https://unpkg.com/@lyrasearch/lyra@0.2.4/dist/browser/src/lyra.js`, it looks like we're bundling using CJS... wouldn't it be the case to build ESM modules for browser usage?

codyzu:
Hmmm... this was related to #120 and was to support webpack 4 builds for expo. Browser seemed like a fitting name even if it was really to support webpack.

I'll see if expo is happy if we just target esmodules

@mateonunez Can you also try changing your import from `browser` to `esm`:
```js
import { create, search, insert } from "https://unpkg.com/@lyrasearch/lyra@0.2.4/dist/esm/src/lyra.js";
```

I'm guessing you just need to run the esm specific build since you running a module in the browser (oh... that is confusing 🤣 )

[...]

PR #138, opened 2022-09-27
docs(readme): document the different builds

codyzu:
* Attempted to make the various builds a bit more clear.
* Added the interesting browser import method (thanks to @mateonunez in #136 )


micheleriva (approved):
That's perfect, thanks Cody 🙏

Issue #139, opened 2022-09-28
Can't search numbers or booleans


mateonunez:
**Describe the bug**
As defined in the type, is not possible to search terms different than `strings`.

![image](https://user-images.githubusercontent.com/11861080/192768058-2ff7066c-68cb-444d-a3a4-b56e855b65ea.png)

**To Reproduce**

```js
import {create, insert, search} from "@lyrasearch/lyra"

(() => {
  const lyra = create({
    schema: {
      author: "string",
      website: "string",
      contributors: "number",
      stars: "number",
      forks: "number",
      language: "string"
    }
  })

  insert(lyra, {
    author: "Lyra",
    website: "https://github.com/LyraSearch",
    contributors: 21,
    stars: 3350,
    forks: 62,
    language: "TypeScript"
  })

  const {hits} = search(lyra, { term: 62 })
})()
```

**Expected behavior**
Lyra's searches should support the types of the `PropertiesSchema` declared.

Anyway, whether I try to cast "**62**" _(forks property)_ into a string, the `hits` are empty.

```js
import {create, insert, search} from "@lyrasearch/lyra"

(() => {
  ...

  const {hits} = search(lyra, { term: "62" })
  console.log({hits}) // empty
})()
```

**Desktop (please complete the following information):**
 - Lyra `v0.2.4`
 - Node `v18.9.0`
 
UPDATED: **Additional context**
Playing with `node_modules` and the build of Lyra, changing the type of term is working fine:

**Boolean**
![image](https://user-images.githubusercontent.com/11861080/192773092-158ff1e8-9cd0-4525-995c-64f026991725.png)

**Number**
![image](https://user-images.githubusercontent.com/11861080/192773234-bad07f38-2b3c-4185-992a-4f8e0d18ae17.png)

micheleriva:
Released in `v0.4.9`

PR #140, opened 2022-09-29
feat(lyra): implement numbers and booleans as searchable

mateonunez:
This PR implements #139.

Often we added `numbers` and `booleans` in Lyra's schema. Those columns should be searchable. Non-string nodes were not inserted into the Trie.

https://github.com/LyraSearch/lyra/blob/b1bf7a6080cc8177624122576a7de23b3c8117b1/src/lyra.ts#L225

Now that nodes are inserted but the tokens are not tokenized in the `recursiveTrieInsertion` method.

```js
{
  '60440268-3': {
    id: '60440268-3',
    key: 't',
    word: 't',
    parent: '60440268-0',
    children: { r: '60440268-4' },
    docs: [],
    end: false
  },
  '60440268-4': {
    id: '60440268-4',
    key: 'r',
    word: 'tr',
    parent: '60440268-3',
    children: { u: '60440268-5' },
    docs: [],
    end: false
  },
  '60440268-5': {
    id: '60440268-5',
    key: 'u',
    word: 'tru',
    parent: '60440268-4',
    children: { e: '60440268-6' },
    docs: [],
    end: false
  },
  '60440268-6': {
    id: '60440268-6',
    key: 'e',
    word: 'true',
    parent: '60440268-5',
    children: {},
    docs: [ '60440268-2' ],
    end: true
  },
  '60440268-7': {
    id: '60440268-7',
    key: '4',
    word: '4',
    parent: '60440268-1',
    children: {},
    docs: [ '60440268-2' ],
    end: true
  }
}
```

I also added a check to not perform the **Levenshtein distance** when the searched term is _non-string_.

The type of the term has been updated allowing the same types supported by the schema.

```bash
-  term: string;
+  term: string | number | boolean;
```

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -24,8 +24,8 @@ export function CANT_DELETE_DOC_NOT_FOUND(id: string): string {
   return `Document with ID ${id} does not exist.`;
 }
 
-export function CANT_DELETE_DOCUMENT(docID: string, key: string, token: string): string {
-  return `Unable to delete document "${docID}" from index "${key}" on word "${token}".`;
+export function CANT_DELETE_DOCUMENT(docID: string, key: string, token: string | number | boolean): string {
+  return `Unable to delete document "${docID}" from index "${key}" on word "${token.toString()}".`;
 }
 
 export function UNSUPPORTED_NESTED_PROPERTIES(): string {
--- a/src/lyra.ts
+++ b/src/lyra.ts
@@ -89,7 +89,7 @@ export type SearchParams<S extends PropertiesSchema> = {
   /**
    * The word to search.
    */
-  term: string;
+  term: string | number | boolean;
   /**
    * The properties of the document to search in.
    */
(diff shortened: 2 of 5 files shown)


micheleriva:
This is quite difficult to deliver at the current stage as it would require a totally different data structure; if we need to search through numbers and booleans as strings, we can just put them inside of the index as string properties:

```js
create({
  schema: {
    myNumber: 'string',
    isFavorite: 'string'
    text: 'string'
  }
})
```

that will do the trick already, as it forces you to convert any boolean/number into a string before inserting it into Lyra.

The reason why we haven't implemented yet numbers/boolean search (as primitives) is that as for now there's just one big data structure (prefix tree) which of course is not a good choice for storing booleans (might be for numbers if converted to radix tree, but again, depends on the query).

The idea is to add more data structures and choose the right one depending on the query you're making, especially once we introduce `AND` and `OR` queries where booleans and numbers are quite important data.

I hope that makes at least little sense

mateonunez:
The implementation of new data structures for primitives sounds good! 

I'm gonna close this pull request looking forward to the new implementations.

PR #157, opened 2022-10-16
perf(lyra): support multiple nested props

### LINKED ISSUE #154: Lyra don't creates nodes for multiple nested props
**Describe the bug**
With #17 Lyra implemented the support of _nested properties_. The implementation works fine if there is only 1 nested property in the schema. Adding more objects in the schema, the nodes are not inserted properly at the root level.

**To Reproduce**

1. Create a new Lyra project and paste the following code:

```js
import { create, insert, search } from "@lyrasearch/lyra";

const lyra = create({
  schema: {
    email: 'string',
    name: { first: 'string', last: 'string' },
    login: { username: 'string', password: 'string' },
  },
})

insert(lyra, {
  email: 'one@two.three',
  name: { first: 'four', last: 'five' },
  login: { username: 'six', password: 'seven' },
})

const results = search(lyra, {
  term: 'six'
})

console.log(results)
```

**Expected behavior**
Everything looks good when searching for the values `four` and `five`, but the searching value belongs to the second block of nested properties (`login`), the results are empty.

**Searching `four` as a term**:
```js
{
  hits: [
    {
      id: '70308384-5',
      email: 'one@two.three',
      name: [Object], // Found here
      login: [Object]
    }
  ],
  term: 'four'
}
```

**Searching `six` as a term**:

```js
{ hits: [], term: 'six' }
```

**Screenshots**

These are the last elements of the `nodes` prop:

image

 - Node **v18.10.0**
 - Lyra **v0.2.6**

mateonunez:
This PR closes #154.

The `recursiveTrieInsertion` method invokes a `return` statement immediately after the call of the self-recursion method.

This impacts not only the objects, but also all the properties that you want to insert after the first object is inserted.

image

By removing the `return', the procedure will consider the other properties following the right insertion strategy.

### DIFF
--- a/src/lyra.ts
+++ b/src/lyra.ts
@@ -221,8 +221,6 @@ function recursiveTrieInsertion<S extends PropertiesSchema>(
     const propName = `${prefix}${key}`;
     if (isNested) {
       recursiveTrieInsertion(index, nodes, doc[key] as ResolveSchema<S>, id, config, propName + ".", tokenizerConfig);
-
-      return;
     }
 
     if (typeof doc[key] === "string") {
--- a/tests/lyra.test.ts
+++ b/tests/lyra.test.ts
@@ -111,7 +111,7 @@ t.test("checkInsertDocSchema", t => {
 });
 
 t.test("lyra", t => {
-  t.plan(16);
+  t.plan(17);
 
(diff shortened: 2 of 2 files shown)


micheleriva (approved):
thank you so much, LGTM!

Issue #159, opened 2022-10-17
Cannot filter strings with more than one space


matteoscaramuzza:
When I try to filter an array of elements like this one:

```json
{
    "ID": 1,
    "Address": "VIA SANTA MARIA 107",
    "PostalCode": "36030",
    "City": "SARCEDO"
}
```
by its address, it doesn't return anything when I try to search by a string with more than 1 space.

e.g. 

![Senza nome](https://user-images.githubusercontent.com/78930980/196231375-ca0054c3-5874-482a-bdcf-5542ff65f087.png)

![Senza nome](https://user-images.githubusercontent.com/78930980/196232368-f5345d69-8cc1-4987-a549-984a5d8bc31a.png)

matteoscaramuzza:
fixed by pr #160

PR #160, opened 2022-10-18
Fixed intersectMany function

matteoscaramuzza:
This change fixes issue #159.
The previous intersectMany method was not finding the correct occurrences.


matteoscaramuzza:
Done 👍🏻

Let me know if you need anything else.

Grazie!

micheleriva (changes requested):
Left a couple of comments, could you please add tests on array intersection specifically?
Something like:

```js
const arr1 = [1,2,3]
const arr2 = [2,3,4]

const intersection = intersectMany(arr1, arr2) // 2,3
```

thanks!

matteoscaramuzza:
I've just added the requested test.
I've also create the type alias and put it inside the datasets folder.

thanks!

tests/datasets/event-json.ts
```diff
@@ -0,0 +1,11 @@
+export type EventJson = {
```
micheleriva, line 1: A couple of things:

1) Type definitions should be inside *.d.ts files, as a general convention
2) This type alias is not used in other script files so I'd just put this at the top of the only test file where it's used, if really needed

tests/utils.test.ts
```diff
@@ -1,8 +1,21 @@
...
+  t.plan(3);
+
+  t.test("should correctly intersect 2 or more arrays", t => {
+    t.plan(4);
+
+    const arr1 = [1, 2, 3, 4, 5, 8];
+    const arr2 = [2, 3, 8];
+    const arr3 = [4, 6, 5, 8];
+
+    t.equal(intersectMany([arr1, arr2]).length, 3);
```
micheleriva, line 14: I wouldn't test length but the actual resulting arrays instead

micheleriva (changes requested):
I left a couple of comments, remember to run the linter before committing 🙂

Thanks again!

tests/datasets/event-json.ts
matteoscaramuzza: Done 👍🏻

micheleriva (approved):
LGTM thanks!

PR #166, opened 2022-10-27
chore: refactor

ilteoood:
Small refactors while I was reading the code

@micheleriva

### DIFF
--- a/src/tokenizer/index.ts
+++ b/src/tokenizer/index.ts
@@ -44,29 +44,28 @@ function normalizeToken(token: string, language: Language, tokenizerConfig: Toke
 
   if (normalizationCache.has(key)) {
     return normalizationCache.get(key)!;
-  } else {
-    // Check if stop-words removal is enabled
-    if (tokenizerConfig?.enableStopWords) {
-      // Remove stop-words
-      if ((tokenizerConfig?.customStopWords as string[]).includes(token)) {
-        const token = "";
-        normalizationCache.set(key, token);
-        return token;
-      }
-    }
-
-    // Check if stemming is enabled
-    if (tokenizerConfig?.enableStemming) {
-      // Stem token when a stemming function is available
-      if (typeof tokenizerConfig?.stemmingFn === "function") {
-        token = tokenizerConfig?.stemmingFn(token);
-      }
-    }
-
-    token = replaceDiacritics(token);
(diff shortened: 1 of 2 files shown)


src/tokenizer/index.ts
```diff
@@ -97,7 +96,7 @@ export function tokenize(
 }
 
 function trim(text: string[]): string[] {
-  while (text[text.length - 1] === "") {
+  while (text.at(-1) === "") {
```
RafaelGSS, line 99: `String.prototype.at(-1)` is slow.

RafaelGSS (changes requested):

src/tokenizer/index.ts
RafaelGSS: https://github.com/RafaelGSS/nodejs-bench-operations/blob/main/RESULTS-v18.md#get-the-last-item-of-an-array

micheleriva: Yeah I guess direct pointer access will always be faster

ilteoood: I didn't knew this. Thanks for sharing the benchmark, I'll use it as a bible 😄

src/tokenizer/index.ts
```diff
@@ -44,29 +44,28 @@ function normalizeToken(token: string, language: Language, tokenizerConfig: Toke
...
-      if ((tokenizerConfig?.customStopWords as string[]).includes(token)) {
-        const token = "";
-        normalizationCache.set(key, token);
-        return token;
-      }
+  }
+  // Check if stop-words removal is enabled
+  if (tokenizerConfig?.enableStopWords) {
+    // Remove stop-words
+    if ((tokenizerConfig?.customStopWords as string[]).includes(token)) {
```
RafaelGSS, line 51: By the way, `Array.prototype.includes` is slow too. But it's out of the topic of this PR.

RafaelGSS (approved):
LGTM.

src/tokenizer/index.ts
micheleriva: @RafaelGSS true, but the the`customStopWords ` array is typically pretty small, so it wouldn't affect the performance that much

micheleriva (approved):
LGTM thanks!

PR #167, opened 2022-10-31
chore: refactor

ilteoood:
Hi, in this PR I've made some refactor about:
- raw comparison instead of Array.includes
- useless else statements
- array creation
- for-of and for-in loops

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -53,7 +53,7 @@ export function uniqueId(): string {
 
 export const reservedPropertyNames = ["id"];
 
-export function getOwnProperty<T = unknown>(object: any, property: string): T | undefined {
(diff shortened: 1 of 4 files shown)


src/utils.ts
```diff
@@ -99,3 +99,12 @@ export function intersectMany<T>(arrays: T[][]): T[] {
     return count === arrays.length;
   });
 }
+
+export const includes = <T>(array: T[] | readonly T[], element: T): boolean => {
```
micheleriva, line 103: If we want to use the `includes` function instead of the built-in `Array.includes` method, I'd recommend turning this into a binary search function. Also, please export a function instead of a constant (for code style consistency)

micheleriva (approved):
lgtm

PR #169, opened 2022-11-05
Feat: adds TF-IDF

micheleriva:
This PR is a rewrite of #149. More complete info to come as soon as out of draft status.

Stages:

- [x] Preload term frequencies in documents
- [x] Calculate TF-IDF scores on search
- [x] Remove frequencies on document deletion
- [x] ~Calculate cosine similarity on TF-IDF-based scoring vectors~ to be done in the future


src/prefix-tree/node.ts
```diff
@@ -32,7 +31,7 @@ export function updateParent(node: Node, parent: Node): void {
 }
 
 export function removeDocument(node: Node, docID: string): boolean {
-  const index = node.docs.indexOf(docID);
+  const index = node.docs.findIndex(doc => doc === docID);
```
ShogunPanda, line 34: Why did you change this?

src/utils.ts
```diff
@@ -91,13 +106,45 @@ export function intersectMany<T>(arrays: T[][]): T[] {
...
-    return count === arrays.length;
-  });
+  }
+
+  return result;
+}
+
+export function insertSortedValue<T = unknown>(arr: T[], el: T, compareFn?: (a: T, b: T) => number): T[] {
+  if (!compareFn) {
+    compareFn = (a, b) => {
```
ShogunPanda, line 122: To allow inlining, extract this function somewhere and then reference it, rather than creating it every time.

ShogunPanda: Moreover, given how you use compareFn, you can insanely simplify it to:

```
function compareFn(a, b) {
  return a - b;
}
```

src/utils.ts
```diff
@@ -107,4 +154,13 @@ export function includes<T>(array: T[] | readonly T[], element: T): boolean {
     }
   }
   return false;
-}
\ No newline at end of file
+}
+
+export function sortTokenScorePredicate(a: TokenScore, b: TokenScore): number {
+  if (a[1] > b[1]) {
```
ShogunPanda, line 160: Same here: you can simplify to `return a[1] - b[1]`.

src/utils.ts
```diff
@@ -71,17 +85,18 @@ export function intersectMany<T>(arrays: T[][]): T[] {
     }
   }
 
-  const set = new Map();
+  const set: Map<string, [number, number]> = new Map();
   for (const elem of arrays[0]) {
-    set.set(elem, 1);
+    set.set(elem[0] ?? "", [1, elem[1]]);
```
ShogunPanda, line 90: I think this is not needed. Since the elem is in the array, how can it be `undefined`?
Same applies on line 97 and 99: the `?? ""` are not needed.

ShogunPanda (commented):
Minor nits, but the rest looks great!

src/prefix-tree/node.ts
micheleriva: Ups, I used to use an object, then switched back to plain string and left that behind. Thanks for noticing!

src/utils.ts
micheleriva: Awesome thanks!

src/utils.ts
micheleriva: Let me try :)

Issue #171, opened 2022-11-06
Implement non-searchable fields in Lyra


micheleriva:
**Is your feature request related to a problem? Please describe.**
This is an RFC for a new Lyra feature, following #168 issue

**Describe the solution you'd like**

**TL;DR**: if a property is not part of the schema definition, don't create an index for it but keep it in the `Lyra.docs` hashmap.

**The Proposal**

We would like to implement non-searchable fields in Lyra.
Right now, we only support three indexable types: `string`, `number`, `boolean`. As for now (`v0.2.8` at the time of writing), creates an index for `string` type only, leaving the `number` and `boolean` indexes to be implemented in future iterations.

My proposal is to make Lyra semi-schemaless, meaning, we only need to define a schema for indexable types.
Let's see the following example. We have the following shape:

```js
[
  {
     "quote": "Patience is the key to joy",
     "author": "Rumi",
     "tags": ["inspirational", "philosophical", "spiritual"],
  }
  ...
]
```

Lyra does not currently support `string[]` types, so we will need to `join()` the array into a single string, such that we'll end up with the following schema definition: 

```js
import { create } from '@lyrasearch/lyra'

create({
  schema: {
    quote: 'string',
    author: 'string',
    tags: 'string'
  }
})
```

This could be fine as Lyra is capable of tokenizing comma-separated strings, but what if we don't want to index the `tags` property?

As for now, we're forced to index the `tags` property anyway, which will force Lyra to create a new tree duplicating a lot of useless data.

The proposal consists into defining the schema of **indexable** properties only, such that given the following data shape:

```js
[
  {
     "quote": "Patience is the key to joy",
     "author": "Rumi",
     "tags": ["inspirational", "philosophical", "spiritual"],
  }
  ...
]
```

We can write down the following schema definition:

```diff
import { create } from '@lyrasearch/lyra'

create({
  schema: {
    quote: 'string',
    author: 'string',
-    tags: 'string'
  }
})
```

Lyra will keep the entire document in the `Lyra.docs` hashmap, but won't tokenize, stem, nor create a tree for the `tags` property, which will be **unsearchable**. 

**It will only be possible to search through the properties defined in the Lyra schema definition**.

As a first iteration, index definitions are immutable and it won't be possible to add a new schema property after Lyra's initialization.

Open to comments and discussions, will be part of Lyra v0.3.0 🙂

ShogunPanda:
I'm totally fine with this proposal.

[...]

PR #174, opened 2022-11-12
Initial non searchable fields support on missing schema/doc key match

LBRDan:
Hello there!
Thanks for this amazing piece of open source software! It should be an example for everyone out there as it is for me ❤️ 

First of all, let me apologize for being so rushy in opening this PR as a draft, but I'm currently running out of battery (hopefully be able to work again on this tonight or tomorrow (CET))
This PR contains a first implementation attempt for #171 issue. I wonder if this is the desired way of implementing this feature, it would be awesome to hear feedback from you

Thanks!

--- Edit:

This PR tries to implement the #171 request for non-searchable fields when they are included inside a document but they're not included inside the schema definition
I tried to implement the feature by:
- Skipping the schema - doc property check adherence inside the `recursiveCheckDocSchema`  function
- Instructing `recursiveTrieInsertion` to be schema-aware when including a document field as searchable by traversing the doc and the schema at the same pace


micheleriva:
Hey hey thank you so much for opening this!
Looking forward to see the PR completed 🙂

Let me know if you need any help at any point!

LBRDan:
> Let me know if you need any help at any point!

Thanks @micheleriva ❤️ 
The implementation is pretty much done... but I think that I'm missing something relative to the nested schema properties
I hope that it's only me not getting it right...

Basically, I don't think that `recursiveCheckDocSchema` is well... actually checking for schema adherence for nested schema/doc properties 😅 
Let me use an example trying to explain this better (using some docs found in the tests)

ATM  `assertDocSchema` uses `recursiveCheckDocSchema` to check the doc (or docs`[]`) against the schema.
https://github.com/LyraSearch/lyra/blob/cc94cf31253168600407b8088f29af0bda45a1a5/src/lyra.ts#L289

While I was expecting that `recursiveCheckDocSchema` was recursively called returning the value of its computation, I found out that using a document (and a schema) with nested fields, the properties of nested properties bypass the checking as nothing gets properly returned from the recursive self call (line 199 below)

https://github.com/LyraSearch/lyra/blob/cc94cf31253168600407b8088f29af0bda45a1a5/src/lyra.ts#L187-L206

Something like the following crashes `recursiveTrieInsertion` in the current `main` branch, but `recursiveCheckDocSchema` should throw before that: (btw static type checking works correctly there)

``` ts
    const db = create({
      schema: {
        quote: "string",
        author: {
          name: "string",
          surname: "string",
        },
        tag: {
          name: "string",
          description: "string",
        },
      },
    });

    insert(db, {
      quote: "Be yourself; everyone else is already taken.",
      author: {
        name: "Oscar",
        surname: "Wild",
      },
      tag: {
        name: "inspirational",
        description: "Inspirational quotes",
        // @ts-expect-error test error case
        cadrega: "so... the apple is the cadrega",
      },
    });
```

However, this other doc that have the same problem but `assertDocSchema` throws as expected:

``` ts
    const db = create({
      schema: {
        quote: "string",
        author: {
          name: "string",
          surname: "string",
        },
        tag: {
          name: "string",
          description: "string",
        },
      },
    });

    insert(db, {
      quote: "A room without books is like a body without a soul.",
      author: {
        name: "Marcus",
        surname: "Tullius Cicero",
      },
      // @ts-expect-error test error case
      furio: "Madga, do you adore me? Then you see it's a mutual thing, do ya?",
      tag: {
        name: "books",
        description: "Quotes about books",
      },
    });
```

The same applies on wrong supported scalar schema types and doc type check

Is this intended to work like this?
If not, after this feature, the nested `doc.tag.cadrega` key should not crash the thing and should throw via `assertDocSchema`, but I'll need to change something more

It's not strictly related to this feature but (if you confirm my doubt) it could prevent the way I tried to implement this from working as expected

Also, if all of that turns out to be something real, the interface for the document `insert` function (to support both doc or doc[]) maybe needs to be slightly modified because it does not distinguish from nested `object` type doc property and "the first iteration for an array of docs", passing recursively the root schema when it's needed to pass the nested property schema

I really hope to be wrong and that I'm missing something 🤞 😓

micheleriva:
Hey @LBRDan thank you so much for the analysis!

So, as for the current version of Lyra (`v0.2.8` at the time of writing) `recursiveCheckDocSchema ` **should** throw an error if there's a nested type/property that does not respect the schema.

For instance, given:

```js
import { create, insert } from '@lyrasearch/lyra'

const db = create({
  quote: 'string'
  meta: {
    rating: 'number'
  }
})
```

this **should** throw:

```js
insert({
  quote: 'foo bar baz',
  meta: {
    rating: 10,
    randomProperty: 'foo' // <-- this should throw
  }
})
```

This is certainly a bug and I'd kindly ask you to open a new issue for it, we can fix it separately after merging this PR (it will be released in `v0.3.0` with #171.

Regarding this PR, I'd ask you to add more test cases, also with deeply nested properties and different types.

Thanks a lot!

micheleriva:
Hi @LBRDan!
Do you have any idea when this PR will be ready for review? No hurry, just planning the next release

micheleriva:
@LBRDan I just tested this locally, and it seems to work great! About the typings, we could consider it as a nice-to-have for a future release. Thank you so much for your contribution

micheleriva (approved):

Issue #175, opened 2022-11-13
Alert before inserting unserializables types into a schemaless instance


micheleriva:
Once we get #174 merged, we will be able to insert any kind of valid JSON object in Lyra.
That also means that the following could be a valid document:

```js
{
  "foo": "bar",
  "baz": () => {},
  "quick": 10n
}
```

In the above example, we can clearly see how it will be impossible to serialize functions and BigInt data without applying any kind of normalization.

My proposal would be to either:

1. Spawn an alert when inserting non-serializable data (functions, maps, sets, bigints, etc.)
2. Throw an error when inserting non-serializable data

I'd personally go with option 2, but I am open for comments

stearm:
Hei @micheleriva ! I just wrote a possible solution, updating the `recursiveCheckDocSchema` function to take advantage of the calculus that's already there. I checked "bigint", "symbol", "function", "Array", "Map", "WeakMap", "Set".

This is the updated implementation:
```ts
function recursiveCheckDocSchema<S extends PropertiesSchema>(
  newDoc: ResolveSchema<S>,
  schema: PropertiesSchema,
): { valid: boolean, key?: string } {
  for (const key in newDoc) {
    const propType = typeof newDoc[key];

    if (!(key in schema)) {
      if(["bigint", "symbol", "function"].indexOf(propType) >= 0) {
        return { valid: false, key };
      }
  
      if(propType === "object") {
        if (["Array", "Map", "WeakMap", "Set"].indexOf(newDoc[key].constructor.name) >= 0) {
          return { valid: false, key };
        }

        return recursiveCheckDocSchema(newDoc[key] as ResolveSchema<S>, schema);
      }

      continue;
    }

    if (propType === "object") {
      return recursiveCheckDocSchema(newDoc[key] as ResolveSchema<S>, schema);
    } else if (propType !== schema[key]) {
      return { valid: false };
    }
  }

  return { valid: true };
}
```
I don't like 100% a couple of things:
- checking in the same function both the schema validity and the unserializables type (but maybe it is worth using the same loop to re-do the calculus)
- the function name is not so clear anymore
- I had to update the return values to have the key and throw a different error

The only doubt is how to catch also custom class instances. What do you think?

LBRDan:
Hello there @stearm
[...]

PR #185, opened 2022-11-17
Fix and improve recursiveCheckDocSchema and first `isSerializable` implementation

LBRDan:
Hello there!
The following PR includes some initial implementations in order to build some momentum for #175 
It also implements a fix for the bug about the recursiveCheckDocSchema failing on reporting wrong nested object schema unmatch, found during the development of the schemaless feature for Lyra ([here](https://github.com/LyraSearch/lyra/pull/174#issuecomment-1312693145))

The serializability detection has been exposed by @stearm [here](https://github.com/LyraSearch/lyra/issues/175#issuecomment-1315058377) and freely adapted here as a first implementation. The same detection is extended even on the nested schema object that could trick the current implementation of `recursiveCheckDocSchema`, apart from the other non matching schema field for schemaless doc fields.


stearm (commented):
We are not showing the user which fields are not serializable, don't we care?

LBRDan:
Hey @stearm 
Sounds good to me to implement a new type of error there
I was wondering what @micheleriva thinks about it in the linked issue before implementing it

I did an initial implementation very similar to yours but mine lacks of schema level awareness when emitting the error (eg. unserializable property inside a nested object)

LBRDan:
Superseded by #300 
Closing this

PR #191, opened 2022-11-19
chore: updated to typescript@4.9.3

jkomyno:
I've noticed the `typescript` dependency hadn't been updated in a while.
Updating to `typescript@4.9.3` didn't require any code change in this repository.

I'd expect compilation times to be slightly faster with this new version.


micheleriva (approved):
that looks good to me. Thanks a lot!

PR #193, opened 2022-11-20
fix(lyra): prevent NaN values in TF-IDF metric. Close #192

### LINKED ISSUE #192: `search` function can result in TF-IDF score with `NaN`  value
**Describe the bug**

Consider the following code from the `search` function in [`lyra.ts`](https://github.com/LyraSearch/lyra/blob/0a910977c174c59d993618d4c67a2d61617966cd/src/lyra.ts#L651-L660):

```typescript
  const termOccurrences = lyraOccurrencies[term];
  // ...

  // Calculate TF-IDF value for each term, in each document, for each index.
  // Then insert sorted results into orderedTFIDFList.
  const documentIDsLength = documentIDs.length;
  for (let k = 0; k < documentIDsLength; k++) {
    const id = documentIDs[k];
    const idf = Math.log10(N / termOccurrencies);
    const tfIdf = idf * (lyraFrequencies?.[id]?.[term] ?? 0);
    // ...
  }
```

The following facts hold:
- `termOccurrencies`'s type is inferred as `number`, but it's actually `number | string | { [k: string]: number } | undefined`
- anomalies in `termOccurrencies`' value make `idf` become `NaN` or `Infinity`, with `tfIdf` resulting in a `NaN` value in turn
- when `termOccurrencies` is a string, object, or undefined -> `idf` becomes `NaN` -> `tfIdf` becomes `NaN`
- when `termOccurrencies` is 0 -> `idf` becomes `Infinity` -> `tfIdf` becomes `NaN`

**To Reproduce**

This problem is best noticeable via the VSCode debugger.

1. Write the following debug configuration in `.vscode/launch.json`:

```json
{
  "version": "0.2.0",
  "configurations": [
    {
      "name": "ts-node",
      "type": "node",
      "request": "launch",
      "args": [
        "${relativeFile}"
      ],
      "runtimeArgs": [
        "-r",
        "ts-node/register"
      ],
      "cwd": "${workspaceRoot}",
      "protocol": "inspector",
      "internalConsoleOptions": "openOnSessionStart"
    }
  ]
}
```

2. Add the following debugger guard to `lyra.ts`, immediately before the `termOccurrencies` definition:

```typescript
  if (Number.isNaN(lyraOccurrencies[term]) || typeof lyraOccurrencies[term] !== "number" || lyraOccurrencies[term] === undefined || lyraOccurrencies[term] === 0) {
    debugger;
  }

  // const termOccurrencies = lyraOccurrencies[term];
```

3. Open `lyra.test.ts` in VSCode, and press `F5`, or go to `Run and Debug` > ▶️ `ts-node` 
4. Look at the stack variables: e.g., `termOccurrencies` is `'function Object() { [native code] }1'` when searching the term `"constructor"` in the `should access own properties exclusively` [test](https://github.com/LyraSearch/lyra/blob/0a910977c174c59d993618d4c67a2d61617966cd/tests/lyra.test.ts#L920-L923).

**Expected behavior**

`termOccurrencies` should always be a `number` larger than 0.

Screenshot 2022-11-20 at 00 40 25

**Desktop:**
 - OS: Mac OS
 - Node: `v18.11.0`

**Additional context**

I've stumbled on this issue while rewriting the [`intersectTokenScores`](https://github.com/LyraSearch/lyra/blob/85f463a3c181da02f057381be316662171ce01e1/src/utils.ts#L75) util in Rust + Wasm.
I already have a fix for this issue, I will push a PR soon.

jkomyno:


src/lyra.ts
```diff
@@ -292,7 +292,8 @@ function recursiveradixInsertion<S extends PropertiesSchema>(
           tokenOccurrencies[propName][token] = 0;
         }
 
-        tokenOccurrencies[propName][token]++;
+        // increase a token counter that may not yet exist
+        tokenOccurrencies[propName][token] = (tokenOccurrencies[propName][token] ?? 0) + 1;
```
jkomyno, line 296: Consider the following, taken from [lyra.ts](https://github.com/LyraSearch/lyra/blob/0a910977c174c59d993618d4c67a2d61617966cd/src/lyra.ts#L295):

```typescript
const propName = 'prop';
const token = 'doc-id';
const tokenOccurrencies = { [propName]: { } };
```

What happens if we increase a non-existing counter like follows?

```typescript
tokenOccurrencies[propName][token]++;
```

Two things happen:
- `tokenOccurrencies[propName]` becomes `{ 'doc-id': NaN }`
- `NaN` is returned from the postfix increment expression

If we wanted to increment a counter that may not yet exist, we should instead run something like:

```typescript
tokenOccurrencies[propName][token] = (tokenOccurrencies[propName][token] ?? 0) + 1;
```

which:
- sets the counter to 1 if it didn't exist
- increment the counter by 1 otherwise

src/lyra.ts
```diff
@@ -648,15 +649,22 @@ export function search<S extends PropertiesSchema>(
     for (let j = 0; j < tokensLength; j++) {
       const term = tokens[j];
       const documentIDs = getDocumentIDsFromSearch(lyra, { ...params, index, term, exact });
-      const termOccurrencies = lyraOccurrencies[term];
+      
+      // lyraOccurrencies[term] can be undefined, 0, string, or { [k: string]: number }
+      const termOccurrencies = typeof lyraOccurrencies[term] === "number"
+        ? (lyraOccurrencies[term] ?? 0)
+        : 0;
```
jkomyno, line 656: - anomalies in `termOccurrencies`' value make `idf` become `NaN` or `Infinity`, with `tfIdf` resulting in a `NaN` value in turn
- when `termOccurrencies` is a string, object, or undefined -> `idf` becomes `NaN` -> `tfIdf` becomes `NaN`
- when `termOccurrencies` is 0 -> `idf` becomes `Infinity` -> `tfIdf` becomes `NaN`

src/lyra.ts
```diff
@@ -648,15 +649,22 @@ export function search<S extends PropertiesSchema>(
     for (let j = 0; j < tokensLength; j++) {
       const term = tokens[j];
       const documentIDs = getDocumentIDsFromSearch(lyra, { ...params, index, term, exact });
-      const termOccurrencies = lyraOccurrencies[term];
+      
+      // lyraOccurrencies[term] can be undefined, 0, string, or { [k: string]: number }
```
jkomyno, line 653: - `termOccurrencies`'s type is inferred as `number`, but it's actually `number | string | { [k: string]: number } | undefined`

I recommend the maintainers to verify why this inferred type is misleading after this PR gets merged.

src/lyra.ts
```diff
@@ -648,15 +649,22 @@ export function search<S extends PropertiesSchema>(
...
+        : 0;
+
       const orderedTFIDFList: TokenScore[] = [];
 
       // Calculate TF-IDF value for each term, in each document, for each index.
       // Then insert sorted results into orderedTFIDFList.
       const documentIDsLength = documentIDs.length;
       for (let k = 0; k < documentIDsLength; k++) {
+        // idf's denominator is shifted by 1 to avoid division by zero
+        const idf = Math.log10(N / (1 + termOccurrencies));
```
jkomyno, line 665: This is such a common scenario that even wikipedia mentions this "unit shift" trick ([link](https://en.wikipedia.org/wiki/Tf%E2%80%93idf#Inverse_document_frequency_2)):

> If the term is not in the corpus, this will lead to a division-by-zero. It is therefore common to adjust the denominator.

jkomyno: This implies that the TF-IDF scores are now slightly smaller in the snapshot tests.

src/lyra.ts
micheleriva: Damn you're right. Thank you so much!

src/lyra.ts
micheleriva: That's right. Thanks a million for that, we used to add `1` and eventually removed it for testing purposes, but never re-introduced it again

micheleriva (approved):
Looks good to me, thank you so much for these fixes @jkomyno!

PR #195, opened 2022-11-28
feat: vscode dev container support

ilteoood:
In this PR I would like to add support for vscode and codespaces dev containers


micheleriva (approved):
Very nice, thanks!

PR #201, opened 2022-12-06
Add sideEffects to package.json

mrbrianevans:
By adding side effects = false to the package.json, it enables tree shaking on certain platforms. 
For example, see webpack: https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free. 
This of course assumes that the package is completely free of side effects (pure).


micheleriva (approved):
thanks for your contribution!

PR #203, opened 2022-12-07
feat(lyra): adds async aliases

micheleriva:
This PR adds async methods to Lyra.
The final goal is to provide an async interface to all the exposed Lyra methods.

When using an async method, the user can access specific optimizations (such as WASM interop #194), that wouldn't be available with a sync method.

This grants backward compatibility and adds the possibility to write more optimizations and lets the user choose whether to adopt them or not.

This PR should get merged right before #194.

The final developer experience should be as easy as:

```js
import { search } from '@lyrasearch/lyra/promises'

await search(...)
```


src/lyra.ts
```diff
@@ -649,11 +648,9 @@ export function search<S extends PropertiesSchema>(
     for (let j = 0; j < tokensLength; j++) {
       const term = tokens[j];
       const documentIDs = getDocumentIDsFromSearch(lyra, { ...params, index, term, exact });
-      
+
       // lyraOccurrencies[term] can be undefined, 0, string, or { [k: string]: number }
-      const termOccurrencies = typeof lyraOccurrencies[term] === "number"
-        ? (lyraOccurrencies[term] ?? 0)
-        : 0;
+      const termOccurrencies = typeof lyraOccurrencies[term] === "number" ? lyraOccurrencies[term] ?? 0 : 0;
```
marco-ippolito, line 653: I think this is hard to read maybe its better in extended form

micheleriva: That's eslint doing its magic 😂

RafaelGSS: why do you need to chekc if it's a number? I'm thinking this whole row can be replaced by

```js
const termOccurrencies = lyraOccurrencies[term] ?? 0
```

PR #210, opened 2022-12-10
Feat: bun ci

marco-ippolito:
resolves: https://github.com/LyraSearch/lyra/issues/188
I've created this feature to introduce bun ci.
I've had problems with caching so I've removed it for the time being.
Also there isn't an official bun-setup action so I'm not sure about this one.
Succesfull run: https://github.com/marco-ippolito/lyra/actions/runs/3664040620/jobs/6194196641
I'm not sure this is the correct approach.

### DIFF
--- a/.github/workflows/bun-ci.yml
+++ b/.github/workflows/bun-ci.yml
@@ -1,4 +1,4 @@
-name: Tests
+name: 'Bun Tests'
 on:
   push:
     branches:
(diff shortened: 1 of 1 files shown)


marco-ippolito:
same as: https://github.com/LyraSearch/lyra/pull/209 it's using the node installed by default on the machine

PR #216, opened 2022-12-14
feat: Only export as ESM.

ShogunPanda:
This PR changes Lyra to be a ESM only package, since most of our target runtime support ESM by default.
For CJS compatibility, a separate export point is exported which can be used to require Lyra in both the following ways:

```
const { create, insert } = require('@lyrasearch/lyra/cjs')

create()
  .then(db => insert(db, {/* ... */})
  .catch(console.error)
```

**OR**

```
const { requireLyra } = require('@lyrasearch/lyra/cjs')

requireLyra((err, lyra) => {
  if(err) {
    throw err
  }

  const { create, insert } = lyra
  create()
    .then(db => insert(db, {/* ... */})
    .catch(console.error)
})
```

Note that `tokenize` and `formatNanoseconds` are not export anymore in `index.ts` nor `cjs` since they are not async methods.


.swcrc
```diff
@@ -0,0 +1,38 @@
+[
+  {
+    "test": ".*\\.ts$",
+    "env": {
+      "targets": "node >= 16"
```
RafaelGSS, line 5: v14 isn't supported?

src/commonjs.cts
```diff
@@ -0,0 +1,9 @@
+type RequireCallback = (err: Error | undefined, lyra?: unknown) => void;
```
RafaelGSS, line 1: is `cts` a valid extension?!

.swcrc
ShogunPanda: Thanks for spotting this. I imported the configuration from a personal project where I'm a little less gentle.
But we can definitely support Node.js 14 here. At least up to April when it goes EOL.

src/commonjs.cts
ShogunPanda: Apparently yes :)

https://www.typescriptlang.org/docs/handbook/esm-node.html

micheleriva: Yes it is, quite unknown tho

.swcrc
mateonunez: Lyra is not supported by Node v14 since `utils.ts > getOwnProperty()` uses the **Object.hasOwn** method, which is not compatible with Node v14.

ShogunPanda: Oh thanks, didn't know that. I'll add an `engines` property to `package.json` then.

mateonunez: Hey @ShogunPanda, what about replacing the `Object.hasOwn` method so that Lyra can be compatible with Node versions <16?

ShogunPanda: @micheleriva Can you remind me why we did include this? I remember it was a bug fix but I don't remember which. :)

cliffordfajardo: Curious, what is the philosophy on the lowest node version to support determined?
Node 18 became the most recent LTS  (long term support) version in Oct 2022

ShogunPanda: Yes, you are right, 18 is the Active LTS.
But 16 is a Maintenance LTS supported up to next September.

Moreover, we don't strictly follow Node release cycle, we just try to support as many versions as possibile without stagnating or doing dirty hacks.

mateonunez: Hey, @ShogunPanda since https://github.com/LyraSearch/lyra/pull/225 has merged the target could be `node >= 14`.

src/types.ts
```diff
@@ -1,8 +1,6 @@
-import type { Language } from "./tokenizer/languages";
-import type { Hooks } from "./methods/hooks";
-import type { TokenizerConfig } from "./tokenizer";
-import type { IIntersectTokenScores } from "./utils";
-import type { Node } from "./radix-tree/node";
+import { IIntersectTokenScores, Language, TokenizerConfig, TokenScore } from "@lyrasearch/components";
```
micheleriva, line 1: `@lyrasearch/lyra` should always provide tree-shakable default algorithms and configurations; we should not rely on external dependencies. Components will override the default algorithms carried by Lyra

package.json
```diff
@@ -1,12 +1,36 @@
...
+    "./cjs": {
+      "types": "./dist/commonjs.d.cts",
+      "require": "./dist/commonjs.cjs"
+    },
+    "./commonjs": {
+      "types": "./dist/commonjs.d.cts",
+      "require": "./dist/commonjs.cjs"
+    },
+    "./promises": {
+      "types": "./dist/promises/index.d.ts",
```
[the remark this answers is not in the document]
ShogunPanda: We do not. Removing it.

package.json
```diff
@@ -56,34 +86,41 @@
       "path": "cz-conventional-changelog"
     }
   },
+  "dependencies": {
+    "@lyrasearch/components": "^0.0.3"
```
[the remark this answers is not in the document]
ShogunPanda: We do need as dependency exactly in order to provide default ones. Or in other words, components is the package that provides the default one.

src/methods/common.ts
```diff
@@ -1,12 +1,10 @@
...
+import { SUPPORTED_LANGUAGES } from "@lyrasearch/components";
+import type { Lyra, PropertiesSchema, ResolveSchema } from "../types";
 import type { SearchParams } from "./search";
-import { includes } from "../utils";
-import { SUPPORTED_LANGUAGES } from "../tokenizer/languages";
-import * as ERRORS from "../errors";
 
 export function assertSupportedLanguage(language: string) {
-  if (!includes(SUPPORTED_LANGUAGES, language)) {
+  if (!(SUPPORTED_LANGUAGES as unknown as string[]).includes(language)) {
```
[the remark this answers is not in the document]
ShogunPanda: I have to check about this. I think it happened due to transpilation on components. Will review this.

src/utils.ts
```diff
@@ -1,8 +1,4 @@
-import type { TokenScore } from "./types";
-
-export type Runtime = typeof knownRuntimes[number];
-
-export type IIntersectTokenScores = (options: { data: TokenScore[][] }) => { data: TokenScore[] };
+import { TokenScore } from "@lyrasearch/components";
```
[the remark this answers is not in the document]
ShogunPanda: It was shared between the two. I will duplicate the definition in order not to have strong dependency.

src/methods/common.ts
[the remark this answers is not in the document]
ShogunPanda: I checked. This should be moved in components as the supported languages are related to the tokenizer.

.swcrc
ShogunPanda: Noted. Changing it.

micheleriva (approved):
LGTM

Issue #220, opened 2022-12-15
Allow custom id


Zombobot1:
**Is your feature request related to a problem? Please describe.**
To delete a document from lyra I need to save its id somewhere.

**Describe the solution you'd like**
Users can specify which field can be used as id during schema declaration.

**Describe alternatives you've considered**
Iterate over all `docs` to find matching data by its id and return doc's id for removal.

micheleriva:
Hi @Zombobot1,
thank you for your feature request. I think it makes sense, but it would only work if the `id` property is defined in the schema definition.

Let's take the following examples:

```js
import { create } from '@lyrasearch/lyra'

const db = create({
  schema: {
    id: 'string',
    foo: 'string'
  }
})
```

in the above example, we could easily implement a feature to delete a document by `id` property.

We wouldn't be able to do the same for the following schema:

```js
const db = create({
  schema: {
   ___ id: 'string',
    foo: 'string'
  }
})
```

unless you specify which one is the ID field with a new API, for instance:

```js
const db = create({
  schema: {
   ___ id: 'string',
    foo: 'string'
  },
  id: '___id'
})
```

but this would cause inconsistent DX if a user is not interested in having a custom `id` field.

Therefore, I'd personally wait for query filters and clauses to create a unified way of deleting stuff depending on certain conditions (i.e.: `WHERE MyCustomID == '123123'`

what do you think?

Zombobot1:
Yes, I wanted to declare a custom id field in my schema. You're right that some users might omit ids and in this case it is possible to auto generate them. [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Using_IndexedDB#structuring_the_database) has such DX. I think omitting id is an edge case because usually people add fulltext search to a table / collection that is stored in a database. For search on a documentation / blog site id might be page's url. If I am right, lyra's current DX is focused on fulfilling an edge case scenario and can be improved.

Query filters are also a decent solution to my problem. However, I assume they will lead to additional indexing that could have been avoided.

micheleriva:
>  If I am right, lyra's current DX is focused on fulfilling an edge case scenario and can be improved.

Yes, and we highly recommend using Lyra _the immutable way_ when on an edge network. Need to change the data? Just make a new deployment, do not mutate the database directly.

> Query filters are also a decent solution to my problem. However, I assume they will lead to additional indexing that could have been avoided.

[...]

Issue #112, opened 2022-08-31
Allow passing custom IDs and doc updating


acnebs:
**Is your feature request related to a problem? Please describe.**
Hi, I have documents that already have their own unique IDs, and when those docs are updated I want to be able to tell Lyra to update the index with that doc in mind. Currently I can do this like so:

1. `const {id: lyraId} = insert(db, doc)`
2. `lyraMap[MY_DOC_ID] = lyraId`
3. an edit is made to _MY_DOC_ID_
4. `const lyraId = lyraMap[MY_DOC_ID]`
5. `remove(db, lyraId)`
6. `const {id: newLyraId} = insert(db, doc)`
7. `lyraMap[MY_DOC_ID] = newLyraId` 

This is far too involved imo.

**Describe the solution you'd like**
The ability to pass a custom ID into the `insert` function, followed by an `update` function that efficiently updates the index with that record in mind. **Or** just modify the `insert` function to be more of an upsert method, e.g. you can always call something like `insert(db, doc, {id: MY_ID})` and it will do the right thing (based on whether the ID already exists or not). If no ID is passed initially then an ID is generated as is the case now.

micheleriva:
Hey @acnebs,
Thank you for opening this. Lyra is built with immutability in mind, even though it provides a `delete` function (to be used if there's data that REALLY needs to be erased).

We'll definitely have to provide a way to update docs, but I'm thinking of making this available exclusively when the `edge` property is set to `false`.

I'd also take the opportunity to make the `delete` function unavailable when running in `edge` mode to preserve immutability.

cc @ShogunPanda what do you think about that?

ShogunPanda:
In my opinion it depends on which kind of editing are you doing to your document.

If you are modifying indexed fields, then a future `update` method (which essentially is a `remove+insert`) will make the trick.

If you are modifying non indexed fields, I would suggest to use a reference to the entire document rather than passing the actual document to Lyra. This way you can modify other fields and Lyra stays immutable.

That said, @micheleriva, adding support to this is trivial since it will mostly (we just have to detect if already existing) to change this line https://github.com/LyraSearch/lyra/blob/main/src/lyra.ts#L337 to something like:

```
const id = config?.id ?? uniqueId();

if(lyra.docs[id]) {
   throw new Error(`A document with id "${id}" already exists.`);
}
```

willtemperley:
[...]

PR #223, opened 2023-01-02
feat(insert): adds custom ID capabilities

micheleriva:
Highly requested feature, will solve both #220 and #112.

By default, Lyra will check if a field `id` exists in the document we're trying to insert:

```js
insert(db, {
  id: 'foo-0123', // Will use this ID
  user: 'mitch'
})
```

if the `id` field does not exist, it will generate a random one.

We can also override the default `id` field by passing a function to the `insert` configuration:

```js
insert(db, {
  id: 'foo-0123',
  user: 'mitch'
}, {
  id: (doc /* the entire document */) => `${doc.user}.01`
})
```

in that case, the `id` will be `mitch.01`. The `id` configuration function can be either sync or async.

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -67,3 +67,11 @@ export function INVALID_STEMMER_FUNCTION_TYPE(): string {
 export function INVALID_TOKENIZER_FUNCTION(): string {
   return `tokenizer.tokenizerFn must be a function.`;
 }
+
+export function TYPE_ERROR_ID_MUST_BE_STRING(type: string): string {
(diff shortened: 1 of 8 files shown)


ShogunPanda (approved):
LGTM!

vassbence:
In the second example is it required to have the `id` prop on the doc?

micheleriva:
> In the second example is it required to have the id prop on the doc?

@vassbence nope, as it will get overwritten by the function 🙂 it's only needed if defined in the schema definition

Zombobot1:
Looks great!

PR #225, opened 2023-01-03
perf(utils): prevent Object.hasOwn from not being defined

mateonunez:
## Context

This PR would replace the `Object.hasOwn` method by using a simple polyfill. This implementation makes Lyra compatible with **Node v14**.

It would be great for Lyra to be used in as many runtimes as possible.

## CI

If this PR would be merged it would be useful to add the tests on specific Node versions such as v14, v16, and v18.

## Tests

The tests were updated by adding the Unit Test to the `getOwnProperty` util method.


micheleriva (approved):
LGTM

PR #234, opened 2023-01-10
fix: Fix document counts.

ShogunPanda:
See: https://lyrasearch.slack.com/archives/C0445CLH631/p1672748793088799


micheleriva (approved):
Awesome thanks

PR #238, opened 2023-01-11
feat(hooks): add afterRemove 

mateonunez:
## Context

This PR would introduce a new hook: **afterRemove**. The hook is very convenient at the time when you are working with cloned indexes and need to perform actions on the data.

This hook is called when a document is removed passing the `docID` as a reference:

> src/methods/remove.ts
```js
  lyra.docs[docID] = undefined;
  lyra.docsCount--;

  if (lyra.hooks.afterRemove) {
    await hookRunner.call(lyra, lyra.hooks.afterRemove, docID);
  }
```

I was undecided whether to return when the method is called: the ID of the deleted document or the entire Lyra instance, but the latter may present problems.

## Tests

The tests were updated by adding the Unit Test to the `afterRemove` hook.

### DIFF
--- a/README.md
+++ b/README.md
@@ -431,6 +431,38 @@ const db = await create({
 await insertWithHooks(db, { author: "test", quote: "test" });
 ```
 
+### afterRemove hook
+
+The `afterRemove` hook is called after the deletion of a document on the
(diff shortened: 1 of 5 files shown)


mateonunez:
Closing this in favor of #310

PR #241, opened 2023-01-12
perf(algorithms): adds new intersection algorithm

micheleriva:


src/algorithms.ts
```diff
@@ -50,3 +51,38 @@ export function intersectTokenScores(arrays: TokenScore[][]): TokenScore[] {
...
+
+      if (token in tokenMap) {
+        tokenMap[token] += score + 0.5
+      } else {
+        tokenMap[token] = score;
+      }
+    }
+  }
+
+  return Object.entries(tokenMap).sort((a, b) => b[1] - a[1]);
```
micheleriva, line 74: @ShogunPanda what if we adopt sparse array insertion here instead of sorting the resulting array?

src/algorithms.ts
```diff
@@ -50,3 +51,38 @@ export function intersectTokenScores(arrays: TokenScore[][]): TokenScore[] {
...
+  const mapsLength = arrays.length;
+  for (let i = 0; i < mapsLength; i++) {
+    const arr = arrays[i];
+
+    const entriesLength = arr.length;
+    for (let j = 0; j < entriesLength; j++) {
+      const [token, score] = arr[j];
+
+      if (token in tokenMap) {
+        tokenMap[token] += score + 0.5
```
micheleriva, line 67: Should we make this `0.5` weight parametrizable?

ShogunPanda: Yes. This way we give people chances to spot how the score was manipulated.

src/algorithms.ts
ShogunPanda: We could, but then you have to filter the array anyway.

micheleriva: Ok let's re-iterate on this in the future

src/algorithms.ts
micheleriva: alright

src/algorithms.ts
```diff
@@ -50,3 +51,38 @@ export function intersectTokenScores(arrays: TokenScore[][]): TokenScore[] {
...
+export function BM25(
+  tf: number,
+  matchingCount: number,
+  docsCount: number,
+  fieldLength: number,
+  averageFieldLength: number,
+  BM25Params: BM25Params
+) {
+  const { k, b, d } = BM25Params;
+  const idf = Math.log((docsCount - matchingCount + 0.5) / (matchingCount + 0.5));
```
lucaong, line 86: You mistakenly removed a `+ 1` from this line here: https://github.com/lucaong/minisearch/blob/1eb584c749ead8209fe4f18132e3f3a693df395e/src/MiniSearch.ts#L1854

micheleriva: You're right, I didn't bring it back from here: https://github.com/LyraSearch/lyra/blob/main/src/methods/search.ts#L168

It's not covered by tests yet, and it's the reason why the PR is still open (alongside other missing features)

Issue #261, opened 2023-01-25
React example is broken


acnebs:
**Describe the bug**
There is no `App.js` file at all – the React example is broken / non-existent.

**To Reproduce**
1. Look at the React example / try to run the React example

**Expected behavior**
A working React example.

**Context**
Seemingly dropped here and therefore not updated (or included) in the `0.4.x` migration: https://github.com/LyraSearch/lyra/commit/d0f95d789ed80de2f8ebc31eedee5e72c0a27877

PR #267, opened 2023-01-30
feat(examples): restored example with react

bozzelliandrea:
hi there, i've just saw the react example issue #261 , so i restore and fixed the example.

if you are thinking why? because i need an example of lyra on react 😃


acnebs:
Unfortunately I don't think this example works with the > `0.4.x` API, which is now `async`

bozzelliandrea:
Good point @acnebs , actually I hadn't thought of an upgrade of the example but only of a restore just for be able to run and try lyra on react
(i used the same version of the vue example, so i think we can upgrade both, probably with another pr)

xyntechx:
Hi! I've made a Lyra + React example using the latest Lyra version. I'll make a separate PR tomorrow.

PR #270, opened 2023-01-31
fix: Vue example with wrong build script

Adibla:
Relatives paths in Vue example seemed to be wrong. I changed them and now 

**npm run dev** on vue-example folder is ok.


micheleriva (approved):
thanks!

PR #271, opened 2023-02-01
feat: update regex tokenizer (en-du-it)

Adibla:
It should resolve this issue:
[Issue](https://github.com/LyraSearch/lyra/issues/230)
I changed only the languages that used the same pattern (du-en-it). I added three unit tests and it seems correct for me, but I wait for your considerations. Thanks


micheleriva (approved):
thanks! lgtm

Issue #274, opened 2023-02-05
Latin Extended character set is not covered in `tokenizer/diacritics`


bogacg:
While doing some tests originating from another inconvenience discussed in #273 I see that characters from ISO Latin Extended list are not covered in source file `tokenizer/diacritics.ts`

I did some research and [Oracle provides a good table](https://docs.oracle.com/cd/E29584_01/webhelp/mdex_basicDev/src/rbdv_chars_mapping.html) of these characters with their decimal codes.

I think I can send a PR using that table to fix the issue.

micheleriva:
Thank you so much for the PR. Looks fantastic to me

micheleriva:
Released in v0.4.6

bogacg:
@micheleriva sorry for missing tests, I'll be more through next time.

PR #275, opened 2023-02-05
Diacritics charcode list fix including ISO Latin Extended.

bogacg:
This change resolves issue #274 

Using information provided in [this page](https://docs.oracle.com/cd/E29584_01/webhelp/mdex_basicDev/src/rbdv_chars_mapping.html) and some Excel magic, we now fully cover ISO Latin1 Extended A character set for diacritics.


micheleriva:
Hi @bogacg!
would you mind adding some tests? Looks great!

micheleriva:
LGTM

PR #276, opened 2023-02-06
fix: do not use eval to fetch values from nested objects

rahuljayaraman:
I would like to use lyra in a chrome extension. Chrome [does not allow](https://developer.chrome.com/docs/extensions/mv3/sandboxingEval/) eval in chrome extensions though.


micheleriva (approved):
LGTM

PR #279, opened 2023-02-09
feat(dsa): adds numbers indexing support

micheleriva:
This PR aims to add support for efficient number indexing via AVL trees. Might switch to min/max heap later.

The initial design for numeric filters is the following:

```js
import { create, search } from '@lyrasearch/lyra'

const db = create({
  name: 'string',
  age: 'number',
  rating: 'number',
  favoriteNumber: 'number'
})

search(db, {
  term: 'michele',
  where: {
    age: {
      between: [20, 30],
    },
    rating: {
      gte: 4
    },
    favoriteNumber: {
      eq: 7
    }
  }
})
```

In the beginning, all the filters will be considered as an intersection of all the subsets of the results (`AND` operation).
This will be released as experimental in `v0.4.X` and will land stable support in the next major release.

### Todo

- [x] KV-AVL implementation
- [x] AVL as reverse index
- [x] AVL as reverse index for multiple docs
- [x] Search filter interface
- [x] Search filter types
- [x] Search filters algorithm 
- [x] Remove numeric interface

### DIFF
--- a/src/radix-tree/node.ts
+++ b/src/radix-tree/node.ts
@@ -1,19 +1,19 @@
 import type { Nullable } from "../types.js";
 import { uniqueId } from "../utils.js";
 
-export interface Node {
+export interface RadixNode {
   id: string;
   key: string;
   subWord: string;
   parent: Nullable<string>;
-  children: Record<string, Node>;
+  children: Record<string, RadixNode>;
   docs: string[];
   end: boolean;
   word: string;
 }
 
-export function create(end = false, subWord = "", key = ""): Node {
-  const node = {
+export function create(end = false, subWord = "", key = ""): RadixNode {
+  const RadixNode = {
     id: uniqueId(),
     key,
     subWord,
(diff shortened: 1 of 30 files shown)


src/radix-tree/node.ts
```diff
@@ -24,34 +24,34 @@ export function create(end = false, subWord = "", key = ""): Node {
...
-  Object.defineProperty(node, "toJSON", { value: serialize });
-  return node;
+  Object.defineProperty(RadixNode, "toJSON", { value: serialize });
+  return RadixNode;
 }
 
-export function updateParent(node: Node, parent: Node): void {
-  node.parent = parent.id;
-  node.word = parent.word + node.subWord;
+export function updateParent(RadixNode: RadixNode, parent: RadixNode): void {
```
ShogunPanda, line 31: Lol. Wrong `Find/Replace All`, isn't it? I think you want to restore the variables names here XD

src/trees/avl/utils.ts
```diff
@@ -0,0 +1,52 @@
+import type { AVLNode } from './node.js';
+
+export const BALANCE_STATE = {
```
ShogunPanda, line 3: Why can't you use the same values you use in `getBalanceFactor`?

src/radix-tree/node.ts
micheleriva: shoot 😂

src/trees/avl/utils.ts
micheleriva: had to refactor a bit, but I agree let me sort this out :)

src/trees/avl/index.ts
```diff
@@ -0,0 +1,214 @@
...
+}
+
+export function rangeSearch<K, V>(node: AVLNode<K, V>, min: K, max: K): V[] {
+  if (!node) {
+    return [];
+  }
+
+  const result: V[] = [];
+
+  function traverse(node: AVLNode<K, V>) {
```
ShogunPanda, line 143: Since this function only uses result from outside function I would move outside and use `results` as first argument.

src/trees/avl/index.ts
```diff
@@ -0,0 +1,214 @@
...
+
+    const temp = findMin(node.right as AVLNode<K, V>);
+    node.key = temp.key;
+    node.right = remove(node.right as AVLNode<K, V>, temp.key);
+  }
+
+  const balanceFactor = getBalanceFactor(node);
+
+  if (balanceFactor === BALANCE_STATE.UNBALANCED_LEFT) {
+    if (getBalanceFactor(node.left as AVLNode<K, V>) === BALANCE_STATE.BALANCED || getBalanceFactor(node.left as AVLNode<K, V>) === BALANCE_STATE.SLIGHTLY_UNBALANCED_LEFT) {
```
mateonunez, line 91: I think you can put `getBalanceFactor(node.left as AVLNode<K, V>)` into a variable and reuse it.

src/trees/avl/index.ts
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    if (getBalanceFactor(node.left as AVLNode<K, V>) === BALANCE_STATE.SLIGHTLY_UNBALANCED_RIGHT) {
+      node.left = rotateLeft(node.left as AVLNode<K, V>);
+      return rotateRight(node);
+    }
+  }
+
+  if (balanceFactor === BALANCE_STATE.UNBALANCED_RIGHT) {
+    if (getBalanceFactor(node.right as AVLNode<K, V>) === BALANCE_STATE.BALANCED || getBalanceFactor(node.right as AVLNode<K, V>) === BALANCE_STATE.SLIGHTLY_UNBALANCED_RIGHT) {
```
mateonunez, line 102: Same here for: `getBalanceFactor(node.right as AVLNode<K, V>)`

src/trees/avl/index.ts
micheleriva: you see a possible inline optimization here?

ShogunPanda: I hope so! ^^

micheleriva: alright then

PR #282, opened 2023-02-12
feat: added suppport for slovenian language

aldingithub:

PR #289, opened 2023-02-15
feat: add support for Bulgarian language

boian-ivanov:
This PR is based on the same functionality being added for Slovenian (#282). 
As Bulgarian is fully based on the Cyrillic alphabet, it closely resembles the Russian regex, excluding the letter "ё" (yo with diaeresis).

It also orders some of the language arrays/lists alphabetically for better readability.


micheleriva (approved):
LGTM!

boian-ivanov:
@micheleriva I'll also look into making a stemmer for it, but most of the solutions I've found are from 10+ years ago and in other languages, so I'll need to port their logic to make it work with Lyra.

PR #291, opened 2023-02-20
feat: add Bulgarian language stemmer

boian-ivanov:
As mentioned on the previous PR (#289), this PR adds a light stemmer for Bulgarian, based on the Java implementation found in Apache Lucene ([repo](https://github.com/apache/lucene), [stemmer](https://github.com/apache/lucene/blob/main/lucene/analysis/common/src/java/org/apache/lucene/analysis/bg/BulgarianStemmer.java)).

### DIFF
--- a/stemmer/lib/bg.d.ts
+++ b/stemmer/lib/bg.d.ts
@@ -1,0 +1,1 @@
+export declare function stemmer(word: string): string;
--- a/stemmer/lib/bg.js
+++ b/stemmer/lib/bg.js
@@ -1,0 +1,160 @@
+/**
+ * Light Stemmer for Bulgarian.
+ *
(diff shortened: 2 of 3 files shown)


boian-ivanov:
I've also been looking into stop-words and wanted to check if we want to have a custom hardcoded solution like the currently available languages in the project, or if we're ok with importing a standard package like [stopwords-iso](https://github.com/stopwords-iso/stopwords-iso), which has a variant for [Bulgarian stopwords](https://github.com/stopwords-iso/stopwords-bg) (which can also be imported standalone).

boian-ivanov:
@micheleriva Is there anything else required to push this over the line?

micheleriva (approved):
Sorry for the late reply 🙂 LGTM

Issue #292, opened 2023-02-20
Typererror when filtering


karisalo1:
ERROR TypeError: filteredIDs is not iterable (cannot read property null)
    at getWhereFiltersIDs (filters.js:64:39)
    at search.js:44:45
    at Generator.next (<anonymous>)
    at asyncGeneratorStep (asyncToGenerator.js:3:1)
    at _next (asyncToGenerator.js:25:1)
    at asyncToGenerator.js:32:1
    at new ZoneAwarePromise (zone.js:1429:21)
    at asyncToGenerator.js:21:1
    at _search (search.js:180:2)
    at search (search.js:19:5)

const db = await create({
      schema: {
        id: 'string',
        brand: 'string',
        model: 'string',
        year: 'number',
        price: 'number',
        used: 'boolean',
        available: 'boolean',
      },
    });
    const data = [
      {
        id: '3e4e0dc3-7cb9-4a03-9a36-2c66f51e08c9',
        brand: 'Acura',
        year: 2000,
        model: 'TL',
        available: true,
        used: false,
        price: 36899,
      },
      {
        id: '321b5c59-be55-48b9-a3d4-3e1586d1a863',
        brand: 'Dodge',
        year: 2009,
        model: 'Journey',
        available: false,
        used: true,
        price: 12799,
      },
      {
        id: '62123f1e-9f13-474a-a282-8d141c296edf',
        brand: 'Nissan',
        year: 1998,
        model: 'Frontier',
        available: true,
        used: false,
        price: 48840,
      },
      {
        id: '62123f1e-9f13-474a-a282-8d141c296eda',
        brand: 'Nissan',
        year: 1980,
        model: 'Micra',
        available: true,
        used: false,
        price: 1000,
      },
    ];
await insertBatch(db, data);
const result = await search(db, {
term: 'Nissan',
where: { year: { eq: 1970 } }
});
console.log('result', result);

Expected behavior:

If selected filter (year 1970) is not available in data, search should just return zero results. 

Additional context:

Possibly the fix is just to check the empty value of filteredIDs in filter.ts, before pushing it to filtersMap.

mateonunez:
Hey @karisalo1, thanks for reporting this.

Can you provide a reproduction of the error? It would be very helpful in finding the error or the fix. Thanks! 🙏

PR #297, opened 2023-02-27
fix: filteredIDs is not iterable

iShibi:
closes:
- #292


micheleriva (approved):
awesome, thank you so much 🙏

PR #300, opened 2023-02-28
Fix and improve `recursiveCheckDocSchema` and first `isSerializable` implementation

LBRDan:
> This PR is basically the same as #185 - It was forgotten in time and the main branch drifted away in the meantime

Original PR #185 
Hello there!
The following PR includes some initial implementations in order to build some momentum for https://github.com/LyraSearch/lyra/issues/175
It also implements a fix for the bug about the recursiveCheckDocSchema failing on reporting wrong nested object schema mismatch, found during the development of the schemaless feature for Lyra (https://github.com/LyraSearch/lyra/pull/174#issuecomment-1312693145)

The serializability detection has been exposed by @stearm https://github.com/LyraSearch/lyra/issues/175#issuecomment-1315058377 and freely adapted here as a first implementation. The same detection is extended even on the nested schema object that could trick the current implementation of recursiveCheckDocSchema, apart from the other non-matching schema field for schemaless doc fields.


src/consts.ts
```diff
@@ -0,0 +1 @@
+export const SUPPORTED_PROPERTY_TYPES = ["string", "number", "boolean"] as const;
```
micheleriva, line 1: Can you rename this file into `constants`?

micheleriva:
Closed due to refactor that included this change as well

PR #317, opened 2023-03-22
feat(elapsed): refactors elapsed time to return an object

micheleriva:
This PR introduces a small change in the returning value of the `search` function.

`search` used to return the following object:

```js
{
  elapsed: bigint,
  hits: [...],
  counts: number,
  ...
}
```

but the bigint type is not serializable in JSON and causes a lot of trouble.

This PR changes the `search` function behavior by returning an object for the `elapsed` property, with the following shape:

```js
{
  elapsed: {
    raw: 181208,            // type: number
    formatted: `181μs` // type: string
  }
}
```

### DIFF
--- a/packages/orama/src/types.ts
+++ b/packages/orama/src/types.ts
@@ -194,6 +194,11 @@ export type SearchContext = {
   docsIntersection: TokenMap
 }
 
+export type ElapsedTime = {
+  raw: number
+  formatted: string
+}
+
 export type Results = {
   /**
    * The number of all the matched documents.
@@ -206,7 +211,7 @@ export type Results = {
   /**
    * The time taken to search.
    */
-  elapsed: bigint | string
+  elapsed: ElapsedTime
   /**
    * The facets results.
    */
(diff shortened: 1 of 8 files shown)


packages/orama/src/types.ts
```diff
@@ -282,7 +287,7 @@ export interface SimpleComponents<S extends Schema = Schema> {
   validateSchema(doc: Document, schema: S): SyncOrAsyncValue<boolean>
   getDocumentIndexId(doc: Document): SyncOrAsyncValue<string>
   getDocumentProperties(doc: Document, paths: string[]): SyncOrAsyncValue<Record<string, string | number | boolean>>
-  formatElapsedTime(number: bigint): SyncOrAsyncValue<bigint> | SyncOrAsyncValue<string>
+  formatElapsedTime(number: bigint): SyncOrAsyncValue<ElapsedTime> | SyncOrAsyncValue<ElapsedTime>
```
ShogunPanda, line 290: This killed me.

micheleriva: Fixed 😂

ShogunPanda (commented):
LGTM!

PR #342, opened 2023-04-08
perf(orama): Node v14 compatibility

### LINKED ISSUE #341: Orama is not compatible with Node v14
**Describe the bug**
Orama is not compatible with Node v14 since the codebase uses the `hasOwn` method.

**Expected behavior**
Orama should be compatible with Node v14 or older versions since the `utils.ts` makes a check on that method.

**Screenshots**
image

mateonunez:
This PR aims to fix #341, making Orama compatible with Node v14.

I've changed the `hasOwn` check and added the tests for it. Also, I've replaced the `replaceAll` and `at` methods, which are not compatible with that version.

### DIFF
--- a/packages/orama/src/utils.ts
+++ b/packages/orama/src/utils.ts
@@ -11,10 +11,11 @@ const second = BigInt(1e9)
 export const isServer = typeof window === 'undefined'
 
 export function sprintf(template: string, ...args: (string | number)[]): string {
-  return template.replaceAll(
+  return template.replace(
     /%(?:(?<position>\d+)\$)?(?<width>-?\d*\.?\d*)(?<type>[dfs])/g,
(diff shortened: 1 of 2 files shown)


packages/orama/src/utils.ts
```diff
@@ -14,10 +11,11 @@ const second = BigInt(1e9)
 export const isServer = typeof window === 'undefined'
 
 export function sprintf(template: string, ...args: (string | number)[]): string {
-  return template.replaceAll(
+  return template.replace(
```
micheleriva, line 14: What problem did you identify in this?

mateonunez: Oh, the replaceAll method was introduced on Node [v15](https://github.com/nodejs/node/releases/tag/v15.0.0) with the V8 8.5 engine..

PR #346, opened 2023-04-17
fixes #345 docs batchSize usage

### LINKED ISSUE #345: Docs Batch Insertion and batchSize usage
On the [docs](https://github.com/oramasearch/orama/blob/main/packages/docs/pages/usage/insert.mdx#batch-insertion) there is an example of `insertMultiple` with `batchSize`  specified via an object while it should be passed as a third parameter.

I caught it because I wanted to try out the library and I did just a copy / paste but it wasn't working 😅

llooz:


micheleriva (approved):
thanks a lot!

PR #357, opened 2023-04-25
feat: Allow empty search.

ShogunPanda:


micheleriva:
Do you think that it would be possible to perform filtering, pagination (limit/offset), and facets with this API?

ShogunPanda:
Filtering and pagination for sure. Facets I have to check the code once I'm back.

PR #358, opened 2023-04-26
fix: disable stopWords to make the example work again

hiimjako:
Hi there, I found that the example provided in the readme does not work.
Here is a fix :)


micheleriva:
Hi there! I fear this edit is not longer needed as we made a slight change in the APIs and stop-words are disabled by default. Thanks anyway 🙏

PR #362, opened 2023-04-30
feat(orama): adds basic synonyms interface and data structures

micheleriva:
Superseeds #311


micheleriva:
Current implementation will not work with stemming. Will need @ShogunPanda back for that.

micheleriva:
closed due to inactivity

PR #375, opened 2023-05-10
feat:add support for dynamic pages to plugin-astro

### LINKED ISSUE #366: plugin-astro does not include blog articles when creating the database
**Describe the bug**
When the database is created, blog pages are not included in the database hence search does not work there.

**To Reproduce**
1. Create a new Astro project
2. Install `@orama/plugin-astro`
3. Setup the plugin (see configuration below)
4. Run `astro build` to generate the db

This is my current config: https://github.com/Balastrong/astro-orama-test/blob/c7ea40528d4a9644d93cfd3c2f1a077cf1ff5255/astro.config.mjs#L13
```ts
    orama({
      // We can generate more than one DB, with different configurations
      mydb: {
        // Required. Only pages matching this path regex will be indexed
        pathMatcher: /blog.+$/,
      },
    }),
```

**Expected behavior**
Blog pages that match the `pathMatcher` rule are included.

**Additional context**
I created a new project with a fresh Astro install at [Balastrong/astro-orama-test](https://github.com/Balastrong/astro-orama-test/), first commit is Astro and second commit is Orama setup.

I see that only `/about` and `/blog` are indexed, but none of the blog pages.

I added a couple console.logs to [prepareOramaDb](https://github.com/oramasearch/orama/blob/13736927697c9ca064f71e9bd71e8a2a71cb7d31/packages/plugin-astro/src/index.ts#L53) in `node_modules` and this is the result:

```ts
// console.log(pages.map((p)=>p.pathname));
[
  '',
  'about/',
  'blog/',
  'blog/markdown-style-guide/',
  'blog/using-mdx/',
  'blog/first-post/',
  'blog/third-post/',
  'blog/second-post/'
]

// console.log(routes.map((r)=>r.route));[
[ '/', '/rss.xml', '/about', '/blog', '/blog/[...slug]' ]
```

If I understood it correctly, by reading the following lines routes are matched with pages and that's where everything inside `blog/` disappears.

https://github.com/oramasearch/orama/blob/13736927697c9ca064f71e9bd71e8a2a71cb7d31/packages/plugin-astro/src/index.ts#L66-L81

Is there a particular reason why we have this extra filter that is actually filtering out all generated pages?

Balastrong:
This PR fixes #366 

After all the investigation reported in the issue, I noticed that static pages are working fine but dynamic routes are not provided by Astro after the build.

With this PR I grab the path of the dist folder by taking the first route, then I compose the path of all other pages from that.

To validate if it works I added three dynamic pages and a new database with a pattern matcher in the sandbox project. Only the page matching the pattern is indexed, as expected. I also expanded the integration tests and they're still passing.

If this makes some kind of sense I'm happy to discuss it here or on Slack!
I'll be really happy to be able to use Orama on my Astro blog with dynamic pages :D

### DIFF
--- a/packages/plugin-astro/src/index.ts
+++ b/packages/plugin-astro/src/index.ts
@@ -63,20 +63,12 @@ async function prepareOramaDb(
 
   // pathname is usually of the form `some/path/`, while `r.route` usually takes
   // the form `/some/path`. That's why we strip start & end slashes to compare.
+  const basePath = routes[0].distURL?.pathname?.replace(/\/$/, '').split('dist/')[0] + 'dist/'
   const pathsToBeIndexed = pages
     .filter(({ pathname }) => dbConfig.pathMatcher.test(pathname))
     .map(({ pathname }) => ({
       pathname,
-      generatedFilePath: routes.filter(r => {
-        const route = r.route.replace(/(^\/|\/$)/g, '')
-        const pathName = pathname.replace(/(^\/|\/$)/g, '')
-
-        if (dbConfig.caseSensitive) {
-          return route.toLowerCase() === pathName.toLowerCase()
-        }
-
-        return route === pathName
-      })[0]?.distURL?.pathname
+      generatedFilePath: basePath + pathname.replace(/^\//, '') + 'index.html'
     }))
     .filter(({ generatedFilePath }) => !!generatedFilePath)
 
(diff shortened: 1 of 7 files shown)


micheleriva:
@simoneb, @castarco is there any change you can take a look at this? I am not very familiar with Astro

simoneb:
Neither am I but the change looks reasonable to me, it also includes a test and an example and existing tests are still running, so LGTM

PR #376, opened 2023-05-11
feat: Provide stemmers as separate package.

ShogunPanda:
This was longly awaited.


packages/orama/package.json
```diff
@@ -20,11 +20,6 @@
       "types": "./dist/components.d.ts",
       "import": "./dist/components.js",
       "require": "./dist/cjs/components.cjs"
-    },
-    "./stemmers/*": {
```
micheleriva, line 24: That means that we're still exporting ALL the stemmers together right? Can you please generate a single export per each stemmer?

packages/stemmers/package.json
```diff
@@ -0,0 +1,69 @@
+{
+  "name": "@orama/stemmers",
+  "version": "1.0.0-beta.15",
+  "type": "module",
+  "description": "Stemmers for Orama",
+  "sideEffects": false,
+  "main": "./dist/index.cjs",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
```
micheleriva, line 10: That means that we're exporting all the stemmers together, right? Can you please create an export for EACH stemmer here?

micheleriva (commented):
We need to export individual stemmers. With the current implementation, we're gonna bundle all the stemmer together.

Also, documentation, README, and code examples need to reflect this change.

packages/stemmers/package.json
ShogunPanda: We already have it: it's the * export on the next entry.
Are you saying we should not have an index at all?

ShogunPanda:
Ok, will remove the index and update all code example accordingly.

ShogunPanda:
@micheleriva I updated the code. Only individual stemmers can be imported now.
I update the documentation page about stemming. In the README I only found a reference to that page so nothing else to update.
Am I missing anything?

README.md
```diff
@@ -101,26 +102,28 @@ await insert(db, {
...
-  }
+    rating: 4.5,
+  },
 })
 
 await insert(db, {
   name: 'Smart LED Bulb',
-  description: 'Control the lighting in your home with this energy-efficient smart LED bulb, compatible with most smart home systems.',
+  description:
+    'Control the lighting in your home with this energy-efficient smart LED bulb, compatible with most smart home systems.',
```
micheleriva, line 112: Pay attention to linting and prettier actions: here is putting the string on a new line (not needed). Do you happen to have a global prettier/eslint config?

package.json
```diff
@@ -95,6 +95,7 @@
     "node": ">= 16.0.0"
   },
   "dependencies": {
-    "@changesets/cli": "^2.26.0"
+    "@changesets/cli": "^2.26.0",
+    "next": "^13.3.0"
```
micheleriva, line 99: Why is Next.js needed as a top-level dependency?

packages/orama/src/errors.ts
```diff
@@ -7,6 +7,7 @@ const errors = {
   NO_LANGUAGE_WITH_CUSTOM_TOKENIZER: 'Do not pass the language option to create when using a custom tokenizer.',
   LANGUAGE_NOT_SUPPORTED: `Language "%s" is not supported.\nSupported languages are:\n - ${allLanguages}`,
   INVALID_STEMMER_FUNCTION_TYPE: `config.stemmer property must be a function.`,
+  MISSING_STEMMER: `As of version 1.0.0 @orama/orama does not ship non English stemmers by default. To solve this, please explicitly import and specify the "%s" stemmer from the package @orama/stemmers`,
```
micheleriva, line 10: I would also add a link to the documentation here

packages/orama/tests/diacritics.test.ts
```diff
@@ -4,7 +4,7 @@ import { replaceDiacritics } from '../src/components/tokenizer/diacritics.js'
 t.test('Diacritics Replacer', t => {
   t.plan(1)
 
-  t.test('Should replace diacritics', t => {
+  t.test('should replace diacritics', t => {
```
micheleriva, line 7: Is this a linter/prettier action?

micheleriva (commented):
left some comments

README.md
ShogunPanda: I think it was VS Code autolinting. Removed it!

package.json
ShogunPanda: It was one of the attempt while Vercel deployments were failing. Removed as well.

packages/orama/tests/diacritics.test.ts
ShogunPanda: No: my OCD kicked in. Lol. Is that an issue?

packages/orama/src/errors.ts
ShogunPanda: Good point. Done!

packages/orama/tests/diacritics.test.ts
micheleriva: Nope, just wondering

README.md
```diff
@@ -101,26 +102,27 @@ await insert(db, {
...
-  }
+    rating: 4.3,
+  },
 })
 
 await insert(db, {
   name: 'Portable Charger',
-  description: 'Never run out of power on-the-go with this compact and fast-charging portable charger for your devices.',
+  description:
+    'Never run out of power on-the-go with this compact and fast-charging portable charger for your devices.',
```
micheleriva, line 121: This is still on a new line

ShogunPanda: It was a different line. Sob. Fixing it.

ShogunPanda: Done

Issue #367, opened 2023-05-06
Plugin Astro fail to build on Windows


TheZal:
**Describe the bug**
When the build command is executed on Windows the build fail due to an error on the destination path of the assets folder.

**To Reproduce**
1. Create a new Astro project
2. Install @orama/plugin-astro
3. Setup the plugin
4. Run astro build

**Expected behavior**
The build to complete and to create the database asset on the folder inside the project

**Screenshots**
![image](https://user-images.githubusercontent.com/12339188/236629958-c442230e-f5e2-447c-81b0-c8dce8c31a7e.png)

**Desktop**
 - OS: Windows 10
 
**Additional context**
The problem is created by the path that is not right escaped for a Windows system.
I can fix it with a PR if you let me :)

micheleriva:
cc. @castarco, @simoneb

micheleriva:
@TheZal can you try it with `orama v1.0.0-beta.16`?

TheZalRevolt:
Gonna try asap!

TheZal:
Gonna try ASAP! :)

TheZal:
It seems that the issue remains...
![image](https://github.com/oramasearch/orama/assets/12339188/fd767b85-1b2b-458b-b1cf-a47d31e6ba81)
![image](https://github.com/oramasearch/orama/assets/12339188/eb085236-af70-4b3f-a2b4-436560a2bde2)

Balastrong:
If my Windows computer will assist me... otherwise I'll pair with @TheZal one day this week :)

micheleriva:
Awesome, thank you @Balastrong! Write me in the Orama Slack channel, I got something for you to thank you for the contribs 🙂

TheZal:
Absolutely happy to pair!

micheleriva:
Thank you folks 🙏

PR #377, opened 2023-05-12
fix: handle pages that are not generated inside a folder

Balastrong:
It's me again! :)

After installing the plugin on my blog I noticed in some cases astro does not generate pages as `dist/path/to/404/index.html` but directly `dist/404.html` which wasn't covered in #375.

I noticed this happened with `404.astro` as it becomes `404.html` while `a404.astro `becomes `a404/index.html`

image

---

I improved the path handling in the script to detect these cases and properly handle them
As usual, I added a test, this time with a permissive matcher to make sure 404 is handled and no errors are thrown.

### DIFF
--- a/packages/plugin-astro/src/index.ts
+++ b/packages/plugin-astro/src/index.ts
@@ -61,15 +61,19 @@ async function prepareOramaDb(
     }
   })
 
-  // pathname is usually of the form `some/path/`, while `r.route` usually takes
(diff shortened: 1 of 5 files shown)


micheleriva (approved):
lgtm

micheleriva:
Thanks @Balastrong!

Any chance you can give this one a try? https://github.com/oramasearch/orama/issues/367 👀

PR #378, opened 2023-05-16
Store multivalue doc field

### LINKED ISSUE #359: Multivalue doc fields
dont see possibility to correctly indexing and  faceting multivalue fields 
something like actors : ["Clint Eastwood ","Tom Hanks"]
facets values and their counts have to be represented for each actors/fields values  separately

allevo:
This PR allows storing and indexing of a multivalue document field.
The search applies an OR on every field: a document matches if just one element matches.

Example of usage:

```javascript
const db = await create({
  name: 'string[]',
  numbers: 'number[]',
  booleans: 'boolean[]',
})
await insert(db,  {
  name: ['Harry', 'James', 'Potter'],
  numbers: [3, 42, 20],
  booleans: [ true, true, false, true],
})
```

fix #359


packages/orama/src/components/facets.ts
```diff
@@ -56,23 +57,60 @@ export async function getFacets(
...
       if (properties[facet] === 'number') {
         for (const range of (facetsConfig[facet] as NumberFacetDefinition).ranges) {
-          if ((facetValue as number) >= range.from && (facetValue as number) <= range.to) {
-            if (facets[facet].values[`${range.from}-${range.to}`] === undefined) {
-              facets[facet].values[`${range.from}-${range.to}`] = 1
-            } else {
-              facets[facet].values[`${range.from}-${range.to}`]++
+          if (Array.isArray(facetValue)) {
+            const alreadyIncremeted = new Set()
+            for (let i = 0; i < facetValue.length; i++) {
```
micheleriva, line 67: As a general rule, replace

```js
for (let i = 0; i < facetValue.length; i++) {
```

with 

```js
const facetValueLength = facetValue.length
for (let i = 0; i < facetValueLength; i++) {
```

to avoid calculating the length of an array on every iteration

micheleriva:
This is very nice!
Just one question tho: wouldn't it be better if we use explicit types for arrays in the schema definition?

Example:

```js

const db = await create({
  schema: {
    name: 'string',
    hobbies: 'string[]' // or 'Array<string>', or whatever
  }
}) 
```

It might be useful to have this information at search/insert time! WDYT?

packages/orama/src/components/facets.ts
```diff
@@ -56,23 +57,64 @@ export async function getFacets(
...
+              continue
+            }
+
+            facets[facet].values[value] = (facets[facet].values[value] ?? 0) + 1
+
+            alreadyInsertedValues.add(value)
+          }
+          break
+        }
+        case 'number[]': {
```
ShogunPanda, line 98: Nit: for easier readability and maintainability, can you please move the `number` case right above this so they are "grouped"?

packages/orama/src/components/index.ts
```diff
@@ -208,32 +210,60 @@ export async function create(
...
-  await implementation.insertDocumentScoreParameters(index, prop, id, tokens, docsCount)
+  const innerSchemaType = schemaType.slice(0, -2) as SearchableType
 
-  for (const token of tokens) {
-    await implementation.insertTokenScoreParameters(index, prop, id, tokens, token)
-    radixInsert(index.indexes[prop] as RadixNode, token, id)
+  const elements = value as Array<string | number | boolean>
+  const elementsLength = elements.length
+  for (let i = 0; i < elementsLength; i++) {
+    insert(implementation, index, prop, id, elements[i], innerSchemaType, language, tokenizer, docsCount)
```
ShogunPanda, line 266: I guess you can directly use `insertScalar` here, isn't it?

packages/orama/src/methods/insert.ts
```diff
@@ -35,20 +35,35 @@ export async function insert(orama: Orama, doc: Document, language?: string, ski
     const actualType = typeof value
     const expectedType = indexablePropertiesWithTypes[key]
 
+    if (expectedType.endsWith('[]') && Array.isArray(value)) {
+      continue
```
ShogunPanda, line 39: I think we want to throw here, like we do on line 43.

packages/orama/src/methods/insert.ts
```diff
@@ -35,20 +35,35 @@ export async function insert(orama: Orama, doc: Document, language?: string, ski
...
 
     await orama.index.beforeInsert?.(orama.data.index, prop, id, value, language, orama.tokenizer, docsCount)
-    await orama.index.insert(orama.index, orama.data.index, prop, id, value, language, orama.tokenizer, docsCount)
+    await orama.index.insert(
+      orama.index,
+      orama.data.index,
+      prop,
+      id,
+      value,
+      expectedType,
```
ShogunPanda, line 62: Note: remeber to update the `Components` page in the docs package to reflect this.

packages/orama/src/types.ts
```diff
@@ -304,6 +304,7 @@ export interface IIndex<I extends OpaqueIndex = OpaqueIndex> {
     prop: string,
     id: string,
     value: SearchableValue,
+    schemaType: SearchableType,
```
ShogunPanda, line 307: Since you add it here, I would add in `IIndexInsertOrRemoveHookFunction` as well and update relevant docs.

ShogunPanda (changes requested):
I have add several minor things but in general looks good to me.
I see search and facets handled, but I don't see filters. Am I missing anything?

packages/orama/src/components/defaults.ts
```diff
@@ -25,17 +25,35 @@ export async function getDocumentIndexId(doc: Document): Promise<string> {
 
 export async function validateSchema<S extends Schema = Schema>(doc: Document, schema: S): Promise<boolean> {
   for (const [prop, type] of Object.entries(schema)) {
+    const value = doc[prop]
+
+    if (typeof type === 'string' && type.endsWith('[]')) {
```
micheleriva, line 30: @ShogunPanda let me know if you agree with this; checking if `type.endsWith('[]')` is clever, but maybe we could just use a dictionary lookup? I feel like it would be a bit cleaner as we add more types

packages/orama/src/components/defaults.ts
```diff
@@ -25,17 +25,35 @@ export async function getDocumentIndexId(doc: Document): Promise<string> {
...
+      if (!Array.isArray(value)) {
+        return false
+      }
+      const expectedType = type.slice(0, -2)
```
micheleriva, line 34: This can be avoided by using dictionary lookups

packages/orama/src/components/facets.ts
micheleriva: I think @allevo put this here as he's grouping lists after the primitives, isn't that ok?

packages/orama/src/components/facets.ts
```diff
@@ -56,23 +57,64 @@ export async function getFacets(
...
-            if (facets[facet].values[`${range.from}-${range.to}`] === undefined) {
-              facets[facet].values[`${range.from}-${range.to}`] = 1
-            } else {
-              facets[facet].values[`${range.from}-${range.to}`]++
+        : (doc![facet] as SearchableValue)
+
+      const alreadyInsertedValues = new Set()
+      switch (properties[facet]) {
+        case 'number': {
+          for (const range of (facetsConfig[facet] as NumberFacetDefinition).ranges) {
```
micheleriva, line 65: I'm wondering if this part can be abstracted as a standalone function to avoid duplication when used with `number[]` type

packages/orama/src/components/index.ts
```diff
@@ -208,32 +210,60 @@ export async function create(
...
+  index: Index,
+  prop: string,
+  id: string,
+  value: SearchableValue,
+  schemaType: SearchableType,
+  language: string | undefined,
+  tokenizer: Tokenizer,
+  docsCount: number,
+): Promise<void> {
+  if (!schemaType.endsWith('[]')) {
```
micheleriva, line 257: Are we sure we don't fall into the `foo[]` error here?

packages/orama/src/errors.ts
```diff
@@ -21,6 +21,7 @@ const errors = {
   UNKNOWN_INDEX: `Invalid property name "%s". Expected a wildcard string ("*") or array containing one of the following properties: %s`,
   INVALID_BOOST_VALUE: `Boost value must be a number greater than, or less than 0.`,
   INVALID_FILTER_OPERATION: `You can only use one operation per filter, you requested %d.`,
+  SCHEMA_VALIDATION_ERROR: `Invalid document.`,
```
micheleriva, line 24: Maybe a bit more explicative error message?

packages/orama/src/methods/search.ts
```diff
@@ -107,7 +107,10 @@ export async function search(orama: Orama, params: SearchParams, language?: stri
     const propertiesToSearchWithTypes = await orama.index.getSearchablePropertiesWithTypes(index)
 
     propertiesToSearch = await orama.index.getSearchableProperties(index)
-    propertiesToSearch = propertiesToSearch.filter((prop: string) => propertiesToSearchWithTypes[prop] === 'string')
+    propertiesToSearch = propertiesToSearch.filter(
+      (prop: string) =>
+        propertiesToSearchWithTypes[prop] === 'string' || propertiesToSearchWithTypes[prop] === 'string[]',
```
micheleriva, line 112: The naive me would have refactored this into `['string', 'string[]'].includes(propertiesToSearchWithTypes[prop])`, now I am not sure if I like it or not ahah

packages/orama/src/types.ts
```diff
@@ -18,9 +18,9 @@ export interface Schema extends Record<string, SearchableType | Schema> {}
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface Document extends Record<string, SearchableValue | Document | unknown> {}
 
-export type SearchableType = 'string' | 'number' | 'boolean'
+export type SearchableType = 'string' | 'number' | 'boolean' | 'string[]' | 'number[]' | 'boolean[]'
 
-export type SearchableValue = string | number | boolean
+export type SearchableValue = string | number | boolean | Array<string> | Array<number> | Array<boolean>
```
micheleriva, line 23: Possibly a stupid question, why are we using `Array<T>` here and `T[]` in `SearchableType`?

packages/orama/src/components/defaults.ts
ShogunPanda: Yes, you are right. Let's go that route.

packages/orama/src/components/facets.ts
ShogunPanda: I was also kinda thinking that.

packages/orama/src/components/index.ts
ShogunPanda: If I'm not mistaken, it would be rejected in `create` phase.

packages/orama/src/methods/search.ts
ShogunPanda: Even a RegExp :)

packages/orama/src/types.ts
ShogunPanda: Agree. I think the linter should have catched this.

packages/orama/src/methods/search.ts
allevo: or startsWith('string') ?

packages/orama/src/components/index.ts
allevo: Correct, during the creation phase, the schema is validated. Also, `schemaType` is typed. So at compile-time and run-time, we are covered.

packages/orama/src/components/facets.ts
allevo: Done, after the rework, the number and number[] cases are very similar. It makes sense to put them together.

packages/orama/src/methods/insert.ts
allevo: Done

packages/orama/src/methods/insert.ts
allevo: Done

packages/orama/src/types.ts
allevo: Done

packages/orama/src/components/defaults.ts
allevo: Done. I have created two functions in default.ts to return if a type is an array and which is the inner type.

packages/orama/src/components/defaults.ts
allevo: Done

packages/orama/src/methods/search.ts
allevo: Done using `.startsWith('string')`.

packages/orama/src/errors.ts
allevo: Removed unused error

retamozo:
This is very helpful. Thnks @allevo ! BTW, @micheleriva any plans to release these changes?

micheleriva:
@retamozo in the coming week :)

PR #379, opened 2023-05-17
Heterogeneous fixes

allevo:
This PR would:
- delete
  - delete document from the storage when the `remove` function is invoked
  - move `remove` tests into the `remove.test.ts` file
  - rename `Delete`documentation section into `Remove`
- insert
  - validate the inserted document against the schema
  - change `validateSchema` interface to return the path of the invalid property instead of just a boolean
  - better test for validation
- create
  - better tests
- other:
  - forward `*Update` and `*MultipleUpdate` from the component to Orama instance
  - avoid accessing to undefined element of a string
  - handle stopword in term
- TODO


packages/orama/src/components/defaults.ts
```diff
@@ -23,22 +23,30 @@ export async function getDocumentIndexId(doc: Document): Promise<string> {
   return await uniqueId()
 }
 
-export async function validateSchema<S extends Schema = Schema>(doc: Document, schema: S): Promise<boolean> {
+export async function validateSchema<S extends Schema = Schema>(doc: Document, schema: S): Promise<string | undefined> {
   for (const [prop, type] of Object.entries(schema)) {
+    const value = doc[prop]
+
+    if (value == undefined) {
```
micheleriva, line 30: Dumb question, why not `typeof value === 'undefined'`?

packages/orama/src/components/defaults.ts
```diff
@@ -23,22 +23,30 @@ export async function getDocumentIndexId(doc: Document): Promise<string> {
   return await uniqueId()
 }
 
-export async function validateSchema<S extends Schema = Schema>(doc: Document, schema: S): Promise<boolean> {
+export async function validateSchema<S extends Schema = Schema>(doc: Document, schema: S): Promise<string | undefined> {
   for (const [prop, type] of Object.entries(schema)) {
+    const value = doc[prop]
```
micheleriva, line 28: You'd be using `typeof value` at least twice here, so maybe it's worth it to instanciate a new variable `const valueType = typeof value`?

allevo: it is used below in the for code

allevo: done

PR #381, opened 2023-05-18
Rework *Multiple methods

allevo:
This PR introduces those changes:
- `remove` method returns false if the document is not found
- `removeAll` returns the number of the removed documents
- `updateMany` doesn't remove the documents if some of them are invalid


micheleriva (approved):
LGTM

PR #384, opened 2023-05-19
Use `sortBy` parameter to sort the result set

allevo:
Introduce a new component to store `indexId`s in ascending order.
This enables an efficient data structure to sort data on the fly.
Support ascending and descending order.
The documents without the property you are sorting for are always put at the end.
Supported type: string, number, boolean
Nested property supported as well

Usage:
```js
const db = await create({
  schema: { key: 'number' },
  sortSchema: { key: 'number' }
})
const [id1, id2, id3, id4, id5, id6] = await insertMultiple(db, [
  { key: 5 },
  { key: 2 },
  { key: 7 },
  { key: 10 },
  { key: -3 },
  { }
])
const resultAsc = await search(db, {
  sortBy: { property: 'key' }
}) // returns the following doc [ id5, id2, id1, id3, id4, id6 ]
const resultDesc = await search(db, {
  sortBy: { property: 'key', order: 'DESC' }
}) // returns the following doc [ id4, id3, id1, id2, id5, id6 ]
const resultCustomOrder = await search(db, {
  sortBy: (a, b) => a[2].key - b[2].key
})
```

TODO:
- ~~cover nested properties with test~~
- ~~cover string properties with test~~
- ~~cover boolean properties with test~~
- ~~write documentation~~
- ~~serialization / deserialization~~
- ~~enabled flag~~
- ~~custom sort algorithm~~


micheleriva:
Just an idea, what about implementing sorting as a polymorphic property, such as:

```js
  term: '',
  sortBy: (doc1, doc2) => doc1.key > doc2.key
})

  term: '',
  sortBy: {
    property: 'key',
    order: 'ASC'
  }
})
```

Or even

```js
  term: '',
  sortingFn: (...docs) => {...}
})
```

It would be nice to be able to override the default sorting method (now in `packages/orama/src/algorithms.ts`, `prioritizeTokenScores` function).

wdyt?

ShogunPanda:
@micheleriva I haven't reviewed changes yet. I advised @allevo to implement this like a top level component (as you're doing with synonyms) so that people can provide their implementation.

I'll update this comment tomorrow when I look at the code.

micheleriva:
@ShogunPanda yes, that's why I was suggesting something like:

```js
  term: '',
  sortingFn: (...docs) => {...}
})
```

this way, anyone can provide their own implementation, as well as making it easier to interact with the sorting process (you simply implement an interface that is similar to JavaScript's default APIs)

allevo:
We want to deliver the functionality by optimizing the effort of the developer. We also want to minimize the time of the sort.
This PR introduces a new component that stores the data correctly. The sort algorithm is O(n).
So, allowing to specify a function is not so correct. But I also understand the need for a "total" custom sort algorithm.

So we could allow both:
```javascript
  term: '',
  sortBy: {
    key: 'myKey',
    sensitivity: 'base' // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare
  }
})
  term: '',
  sortBy: (doc1, doc2) => doc1.key > doc2.key
})
```

The first one is optimized, but the second one isn't.

NB: the second one allows also the developer to change the stored object. No one can prevent writing something like
```javascript
  term: '',
  sortBy: (doc1, doc2) => {
    doc1.foo = 42
    return doc1.key > doc2.key
  }
})
```
If it happens, the changes will not apply to the index.

So, I go to support both.

micheleriva:
@allevo I agree with your comment. I think we should provide an optimized function out of the box, and give the possibility of providing a custom one if needed.

micheleriva:
@allevo my only concern with something like 

```js
  term: '',
  sortBy: {
    key: 'myKey',
    sensitivity: 'base'
  }
})
```
is: how clear is it to the final user what's going on?

allevo:
uhm...

What do you think about this one?
```javascript
const db = await create({
  schema: { key: 'number' },
  sortSchema: { key: 'number' }
})
  term: '',
  sortBy: 'number', // or '-number'
}) // optimized
  term: '',
  sortBy: (doc1, doc2) => doc1.number > doc2.number
}) // not optimized
```

micheleriva:
I'd make the order more explicit:

```js
  term: '',
  sortBy: {
    number: 'ASC' //
  }
}) // optimized
```

or

```js
  term: '',
  sortBy: {
    property: 'number',
    order: 'ASC'
  }
}) // optimized
```

and also support:

```js
  term: '',
  sortBy: (doc1, doc2) => doc1.number > doc2.number
}) // not optimized
```

I don't think `sortSchema: { key: 'number' }` as sorting should only be available on `string` and `number` types

ShogunPanda:
3 things on the fly:

1. I also like `{sortBy: {key: 'ASC' }}` syntax
2. I thing we should support sorting by string, numbers and booleans. No array type sorting.
3. Allowing function for sortBy would be a deviation in the API compared to other components. I would instead go for `{components: {sorter: ... }}` that implements a new `ISorter` interface. Like is done for other parts.

I'll elaborate more tomorrow.

allevo:
I like this combo:
```javascript
  term: '',
  sortBy: {
    property: 'number',
    order: 'ASC'
  }
}) // optimized
  term: '',
  sortBy: (doc1, doc2) => doc1.number > doc2.number
}) // not optimized
```

In this way, the user cannot specify more than one field to sort on.

@ShogunPanda:
- I agree with you on the 2°
- your 3rd point is really a good one!

ShogunPanda:
@allevo  you are right. Let's go for:

`sortBy: {property: "key", order: "ASC"}` 

and of course sortBy can accept an array for cascade sorting.
Note that cascade sorting might or might not be supported by the `sorter` component (and our default implementation will support it)

packages/docs/pages/usage/search/sort.mdx
```diff
@@ -0,0 +1,58 @@
...
+  term: 'prestige',
+  sort: {
+    property: 'title', // or 'year', 'isTop'
+  }
+})
+```
+
+Orama upports sorting on 'string', 'number' and 'boolean'.
+
+You can also specify nested properties using the '.' notation: 'meta.tag', 'meta.rating' and 'meta.favorite'
```
micheleriva, line 30: Please add backticks:

```
You can also specify nested properties using the `.` notation: `'meta.tag'`, `'meta.rating'` and `'meta.favorite'`
```

packages/docs/pages/usage/search/sort.mdx
```diff
@@ -0,0 +1,58 @@
+# Sort
```
micheleriva, line 1: I would call this page "Sorting"

packages/docs/pages/usage/search/_meta.json
```diff
@@ -3,6 +3,7 @@
   "fields-boosting": "Fields boosting",
   "facets": "Facets",
   "filters": "Filters",
+  "sort": "Sort",
```
micheleriva, line 6: I would call this page "Sorting"

packages/docs/pages/internals/components.mdx
```diff
@@ -274,6 +274,52 @@ const movieDB = await create({
...
+    title: 'string',
+    director: 'string',
+  },
+  sort: {
+    title: 'string',
+    director: 'string',
+  }
+  components: {
+    sort {
+      // Define the custom component here
```
micheleriva, line 317: I would put an example of a custom component here

packages/docs/pages/usage/search/sort.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+To sort, Orama needs to know on which properties you want to sort.
```
micheleriva, line 3: Is this still needed? Given that we can apply sorting on every property (except for booleans), we might not need a `sortSchema`

packages/docs/pages/internals/components.mdx
```diff
@@ -274,6 +274,75 @@ const movieDB = await create({
 })
 ```
 
+### `sort`
```
ShogunPanda, line 277: I know this is gonna be big, but I would call this `sorter`. The verb alone makes little sense.

packages/docs/pages/internals/components.mdx
```diff
@@ -274,6 +274,75 @@ const movieDB = await create({
...
+import { sort as defaultSort } from '@orama/orama/components.js'
+import { DefaultSort, Sort } from '@orama/orama/components/sort.js'
+
+const s = await defaultSort.createSort() as DefaultSort
+const db = await create({
+  schema: {
+    number: 'number'
+  },
+  components: {
+    sort: {
```
ShogunPanda, line 314: I would make this example smaller. For instance, just override a single function and propagate all the others.

packages/docs/pages/usage/search/sorting.mdx
```diff
@@ -0,0 +1,79 @@
...
+})
+const results = await search(db, {
+  term: 'prestige',
+  sort: {
+    property: 'title', // or 'year', 'isTop'
+  }
+})
+```
+
+Orama supports sorting on 'string' and 'number'.
```
ShogunPanda, line 26: Non blocking: in the future we can also support boolean pretty easily I guess once we define the order.

packages/docs/pages/usage/search/sorting.mdx
```diff
@@ -0,0 +1,79 @@
...
+    title: 'string',
+    year: 'number',
+    isTop: 'boolean',
+    meta: {
+      tag: 'string',
+      rating: 'number',
+      favorite: 'boolean',
+    },
+  },
+  sort: {
```
ShogunPanda, line 75: I'm not sure about this. 
It's probably fine now, but we might end up with possible endless list of options in the future.

@micheleriva WDYT about moving to an approach likes this?

```
...
schema: {
  title: 'string',
  year: {type: 'number', sortable: false }
  // ...
},
```

packages/orama/src/methods/create.ts
```diff
@@ -73,7 +77,7 @@ function validateComponents(components: Components) {
   }
 }
 
-export async function create({ schema, language, components, id }: CreateArguments): Promise<Orama> {
+export async function create({ schema, sort: s, language, components, id }: CreateArguments): Promise<Orama> {
```
ShogunPanda, line 80: Please rename `s` with a more meaningful name, like `sortConfig`.

packages/orama/src/components/sort.ts
```diff
@@ -0,0 +1,222 @@
...
+  value: SortValue,
+  schemaType: SortType,
+  language: string | undefined,
+): Promise<void> {
+  const s = sort.sorts[prop] as PropertySort<SortValue>
+
+  let predicate: (value: [string, SortValue]) => boolean
+  switch(schemaType) {
+    case "string":
+      predicate = (d: [string, SortValue]) => (d[1] as string).localeCompare(value as string, language) > 0
```
ShogunPanda, line 102: For performance reason, extract these two functions and use them below using `.bind`.

ShogunPanda (changes requested):
This is awesome.
I left some comments for improvements. Some are pretty obnoxious but it shouldn't take too much to implement them.

packages/docs/pages/usage/search/sorting.mdx
allevo: At this point, why not move to JsonSchema? Creating a "custom" schema language is useless, and the user can confuse it!
Anyway, I prefer to keep it simple: by default, all are sortable; if you want, you can put the properties inside the denied list.

micheleriva: Agreed with @allevo. Orama must be simple to use and customize. You can use it with its out of the box features, and then optimize it when/if needed

packages/docs/pages/usage/search/sorting.mdx
micheleriva: Why should I sort `true` and `false` values?

packages/docs/pages/usage/search/sorting.mdx
```diff
@@ -0,0 +1,93 @@
...
+    meta: {
+      tag: 'string',
+      rating: 'number',
+      favorite: 'boolean',
+    },
+  }
+})
+const results = await search(db, {
+  term: 'prestige',
+  sort: {
```
micheleriva, line 20: Just wondering if `"sort"` is the correct name here. Maybe `"sortBy"`?

packages/docs/pages/usage/search/sorting.mdx
```diff
@@ -0,0 +1,93 @@
...
+    property: 'title', // or 'year', 'isTop'
+  }
+})
+```
+
+Orama supports sorting on 'string' and 'number'.
+
+You can also specify nested properties using the '.' notation: `'meta.tag'`, `'meta.rating'` and `'meta.favorite'`
```
micheleriva, line 28: Please add a simple example for that

packages/docs/pages/usage/search/sorting.mdx
```diff
@@ -0,0 +1,93 @@
...
+      rating: 'number',
+      favorite: 'boolean',
+    },
+  }
+})
+const results = await search(db, {
+  term: 'prestige',
+  sort: {
+    property: 'title', // or 'year', 'isTop'
+    order: 'DESC'
```
micheleriva, line 52: Add a comment here with `// default is "ASC"`

packages/docs/pages/usage/search/sorting.mdx
```diff
@@ -0,0 +1,93 @@
...
+      tag: 'string',
+      rating: 'number',
+      favorite: 'boolean',
+    },
+  }
+})
+const results = await search(db, {
+  term: 'prestige',
+  sort: {
+    property: 'title', // or 'year', 'isTop'
```
micheleriva, line 51: We don't sort booleans, so `isTop` should not appear here

packages/docs/pages/usage/search/sorting.mdx
```diff
@@ -0,0 +1,93 @@
...
+    year: 'number',
+    isTop: 'boolean',
+    meta: {
+      tag: 'string',
+      rating: 'number',
+      favorite: 'boolean',
+    },
+  },
+  sort: {
+    deniedProperties: ['yeay', 'meta.tag']
```
micheleriva, line 76: Typo in `yeay`

packages/orama/src/components/sorter.ts
```diff
@@ -0,0 +1,267 @@
...
+          orderedDocs: [],
+          type: type,
+          n: 0
+        }
+        break
+      case 'boolean':
+      case 'boolean[]':
+      case 'number[]':
+      case 'string[]':
+        // We don't allow to sort by arrays
```
micheleriva, line 72: We should put this in the docs as well

packages/orama/src/components/sorter.ts
```diff
@@ -0,0 +1,267 @@
...
+  if (!sorter.enabled) {
+    return
+  }
+  const s = sorter.sorts[prop] as PropertySort<SortValue>
+
+  const index = s.docs[id]
+  delete s.docs[id]
+
+  // Decrement position for the greather documents
+  for (let i = index + 1; i < s.orderedDocs.length; i++) {
```
micheleriva, line 157: Here you are accessing a nested property and calculating its length for each iteration; move this to a separate constant:

```js
const orderedDocsLength = s.orderedDocs.length
for (let i = index + 1; i < orderedDocsLength; i++) {
```

packages/orama/src/components/sorter.ts
```diff
@@ -0,0 +1,267 @@
...
+  if (index === -1) {
+    index = s.orderedDocs.length
+    s.orderedDocs.push([id, value])
+  } else {
+    s.orderedDocs.splice(index, 0, [id, value])
+  }
+  s.docs[id] = index
+
+  // Increment position for the greather documents
+  for (let i = index + 1; i < s.orderedDocs.length; i++) {
```
micheleriva, line 137: Here you are accessing a nested property and calculating its length for each iteration; move this to a separate constant:

```js
const orderedDocsLength = s.orderedDocs.length
for (let i = index + 1; i < orderedDocsLength; i++) {
```

packages/orama/src/methods/search.ts
```diff
@@ -191,18 +191,21 @@ export async function search(orama: Orama, params: SearchParams, language?: stri
...
     uniqueDocsArray = intersectFilteredIDs(whereFiltersIDs, uniqueDocsArray)
   }
 
-  const resultIDs: Set<string> = new Set()
-  // Populate facets if needed
-  const facets = shouldCalculateFacets ? await getFacets(orama, uniqueDocsArray, params.facets!) : {}
+  if (params.sortBy) {
+    uniqueDocsArray = await orama.sorter.sortBy(orama.data.sorter, uniqueDocsArray, params.sortBy)
+  } else {
+    uniqueDocsArray = uniqueDocsArray.sort(sortTokenScorePredicate)
```
micheleriva, line 205: @allevo is this where users can define their own sorting predicate?

packages/docs/pages/internals/components.mdx
```diff
@@ -274,6 +274,57 @@ const movieDB = await create({
...
+  * `id`: The ID of the document being inserted.
+  * `value`: The value of the property in the document.
+  * `schemaType`: The type of the property in the document according with the sort schema.
+  * `language`: The language of the document.
+* `remove`: A function that removes a document from the index. It receives the following arguments:
+  * `sorter`: The sorter returned by the `create` function.
+  * `prop`: The property that it is currently considered.
+  * `id`: The ID of the document being inserted.
+* `sortBy`: A function that inserts document information into the index for future results score calculation. It should be typically invoked within `insert`. It receives the following arguments:
+  * `sorter`: The index.
```
ShogunPanda, line 295: It should be `The sorter returned by the `create` function.` I think.

packages/docs/pages/internals/components.mdx
```diff
@@ -274,6 +274,57 @@ const movieDB = await create({
...
+
+const s = await defaultSorter.createSorter()
+const db = await create({
+  schema: {
+    number: 'number'
+  },
+  components: {
+    sorter: {
+      // override partially the default sorter
+      ...(s as unknown as ISorter),
```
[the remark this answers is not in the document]
ShogunPanda: In theory if `createSorter` returns a valid type this is unneeded completely IMHO.

packages/docs/pages/usage/search/sorting.mdx
ShogunPanda: For exactly this reason. Lol. You might want to have the docs with some field set to `true` and then `false` true.
Think about having the or `inPromotion` property. In the results, regardless of the score, you might want to have the promoted items first.

packages/docs/pages/usage/search/sorting.mdx
ShogunPanda: See my comment above.

packages/docs/pages/usage/search/sorting.mdx
ShogunPanda: Ok, no problem for me.

packages/docs/pages/usage/search/sorting.mdx
ShogunPanda: Sorry, I'm a nitpicker. `denied` sounds weird here. 
What about `unsortableProperties` or `skipProperties` (which is what we already have in the tokenizer)?

packages/orama/src/components.ts
```diff
@@ -2,3 +2,4 @@ export * from './components/defaults.js'
 export * as documentsStore from './components/documents-store.js'
 export * as index from './components/index.js'
 export * as tokenizer from './components/tokenizer/index.js'
+export * as sort from './components/sorter.js'
```
ShogunPanda, line 5: I guess we want `sorter` here as well.

packages/orama/src/types.ts
```diff
@@ -438,19 +495,26 @@ export type Components = Partial<ObjectComponents & FunctionComponents & SingleO
...
+  Index: OpaqueIndex;
+  DocumentStore: OpaqueDocumentStore;
+  Sorter: OpaqueSorter;
+}>
 
-interface Data<I extends OpaqueIndex, D extends OpaqueDocumentStore> {
+interface Data<I extends OpaqueIndex, D extends OpaqueDocumentStore, S extends OpaqueSorter> {
   index: I
   docs: D
+  sorter: S
```
ShogunPanda, line 508: Here the name does not follow the component it belongs to. For readability I would rename this to `sorting`.

ShogunPanda (changes requested):
Another round of nits. But you're doing a great job!

packages/docs/pages/usage/search/sorting.mdx
micheleriva: The risk here is to make `skipProperties` ambiguous. Maybe just `denyProperties`

ShogunPanda: I would then go for `unsortableProperties` or just `unsortable`.

packages/orama/src/methods/search.ts
allevo: Now it is allowed. The user can specify a custom sort function

packages/orama/src/methods/serialization.ts
```diff
@@ -3,16 +3,19 @@ import { Orama } from '../types.js'
 export interface RawData {
   index: unknown
   docs: unknown
+  sort: unknown
```
ShogunPanda, line 6: Non-blocking: I would probably call this `sorting` since so far we have 1-1 mapping.

micheleriva (approved):
LGTM

PR #389, opened 2023-05-25
feat: rewrites AVL tree removing recursion at all

micheleriva:
This PR is a complete rewrite of the current AVL implementation to avoid recursive calls at all. This will allow Orama to work on larger numbers (and array of numbers) without typical problems linked to recursion.

### DIFF
--- a/packages/orama/src/trees/avl.ts
+++ b/packages/orama/src/trees/avl.ts
@@ -6,6 +6,11 @@ export type Node<K, V> = {
   height: number
 }
 
+type StackNode<K, V> = {
+  node: Node<K, V>,
+  checkedChildren: boolean
+};
+
 const BALANCE_STATE = {
   UNBALANCED_RIGHT: 1,
   SLIGHTLY_UNBALANCED_RIGHT: 2,
@@ -53,29 +58,66 @@ function rotateRight<K, V>(node: Node<K, V>): Node<K, V> {
   return left
 }
 
-function findMin<K, V>(node: Node<K, V>): Node<K, V> {
-  return node.left ? findMin(node.left) : node
-}
-
 export function contains<K, V>(node: Node<K, V>, key: K): boolean {
   return !!find(node, key)
 }
 
-export function getSize<K, V>(node: Node<K, V> | null): number {
-  if (!node) {
-    return 0
+export function getSize<K, V>(root: Node<K, V> | null): number {
+  let size = 0;
+  const queue: Array<Node<K, V>> = [];
+
+  if (root !== null) {
+    queue.push(root);
   }
 
-  return 1 + getSize(node.left) + getSize(node.right)
+  while (queue.length > 0) {
+    const node = queue.shift() as Node<K, V>;
+    size++;
+
+    if (node.left !== null) {
+      queue.push(node.left);
+    }
+
+    if (node.right !== null) {
+      queue.push(node.right);
+    }
+  }
+
+  return size;
 }
 
-export function isBalanced<K, V>(node: Node<K, V> | null): boolean {
-  if (!node) {
-    return true
+export function isBalanced<K, V>(root: Node<K, V> | null): boolean {
+  const stack: StackNode<K, V>[] = [];
+
+  if (root !== null) {
+    stack.push({ node: root, checkedChildren: false });
   }
 
-  const heightDiff = Math.abs(getHeight(node.left) - getHeight(node.right))
-  return heightDiff <= 1 && isBalanced(node.left) && isBalanced(node.right)
+  while (stack.length > 0) {
+    const top = stack[stack.length - 1];
+
+    if (top.checkedChildren) {
+      const heightDiff = Math.abs(getHeight(top.node.left) - getHeight(top.node.right));
+
+      if (heightDiff > 1) {
+        return false;
+      }
+
+      stack.pop();
+    } else {
+      top.checkedChildren = true;
+
+      if (top.node.right !== null) {
+        stack.push({ node: top.node.right, checkedChildren: false });
+      }
+
+      if (top.node.left !== null) {
+        stack.push({ node: top.node.left, checkedChildren: false });
+      }
+    }
+  }
+
+  return true;
 }
 
 export function rangeSearch<K, V>(node: Node<K, V>, min: K, max: K): V {
@@ -172,20 +214,17 @@ export function lessThan<K, V>(node: Node<K, V>, key: K, inclusive = false): V {
   return result as V
 }
 
-function getNodeByKey<K, V>(node: Node<K, V>, key: K): Node<K, V> | null {
-  if (!node) {
-    return null
+function getNodeByKey<K, V>(node: Node<K, V> | null, key: K): Node<K, V> | null {
+  while (node !== null) {
+    if (key < node.key) {
+      node = node.left;
+    } else if (key > node.key) {
+      node = node.right;
+    } else {
+      return node;
+    }
   }
-
-  if (node.key === key) {
-    return node
-  }
-
-  if (key < node.key) {
-    return node.left ? getNodeByKey(node.left, key) : null
-  }
-
-  return node.right ? getNodeByKey(node.right, key) : null
+  return null;
 }
 
 export function create<K, V>(key: K, value: V): Node<K, V> {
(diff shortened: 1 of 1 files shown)


ShogunPanda (approved):

packages/orama/src/trees/avl.ts
```diff
@@ -53,29 +58,66 @@ function rotateRight<K, V>(node: Node<K, V>): Node<K, V> {
   return left
 }
 
-function findMin<K, V>(node: Node<K, V>): Node<K, V> {
-  return node.left ? findMin(node.left) : node
-}
-
 export function contains<K, V>(node: Node<K, V>, key: K): boolean {
   return !!find(node, key)
 }
```
marco-ippolito, line 63: This is just a passthrough function I think can be removed.

packages/orama/src/trees/avl.ts
```diff
@@ -53,29 +58,66 @@ function rotateRight<K, V>(node: Node<K, V>): Node<K, V> {
...
 
-export function getSize<K, V>(node: Node<K, V> | null): number {
-  if (!node) {
-    return 0
+export function getSize<K, V>(root: Node<K, V> | null): number {
+  let size = 0;
+  const queue: Array<Node<K, V>> = [];
+
+  if (root !== null) {
```
marco-ippolito, line 69: Is there a particular reason why this value set it to null instead of undefined?

packages/orama/src/trees/avl.ts
```diff
@@ -198,121 +237,161 @@ export function create<K, V>(key: K, value: V): Node<K, V> {
...
+      node = node.right as Node<K, V>;
+    } else {
+      return node.value;
     }
+  }
 
-    if (!node.left) {
-      return node.right as Node<K, V>
-    }
+  return null;
```
allevo, line 333: should we use getNodeByKey here ?

micheleriva: Not sure I follow?

packages/orama/src/trees/avl.ts
```diff
@@ -53,29 +58,66 @@ function rotateRight<K, V>(node: Node<K, V>): Node<K, V> {
...
+    stack.push({ node: root, checkedChildren: false });
   }
 
-  const heightDiff = Math.abs(getHeight(node.left) - getHeight(node.right))
-  return heightDiff <= 1 && isBalanced(node.left) && isBalanced(node.right)
+  while (stack.length > 0) {
+    const top = stack[stack.length - 1];
+
+    if (top.checkedChildren) {
+      const heightDiff = Math.abs(getHeight(top.node.left) - getHeight(top.node.right));
```
marco-ippolito, line 100: I think it might be possible to skip the math.abs and optimize it a bit and just check if greater than 1 or smaller than -1 I need to think a bit more about this

packages/orama/src/trees/avl.ts
```diff
@@ -172,20 +214,17 @@ export function lessThan<K, V>(node: Node<K, V>, key: K, inclusive = false): V {
...
+      node = node.left;
+    } else if (key > node.key) {
+      node = node.right;
+    } else {
+      return node;
+    }
   }
-
-  return node.right ? getNodeByKey(node.right, key) : null
+  return null;
```
marco-ippolito, line 227: This is not idomatic of javascript, unless a specific case its better to just return

packages/orama/src/trees/avl.ts
```diff
@@ -198,121 +237,161 @@ export function create<K, V>(key: K, value: V): Node<K, V> {
   }
 }
 
-export function insert<K, V>(node: Node<K, V>, key: K, value: V): Node<K, V> {
-  if (!node) {
-    return create(key, value)
+export function insert<K, V>(root: Node<K, V>, key: K, value: V): Node<K, V> {
+  let parent = null;
```
marco-ippolito, line 241: Same here probably no need to initialize it.

packages/orama/src/trees/avl.ts
```diff
@@ -198,121 +237,161 @@ export function create<K, V>(key: K, value: V): Node<K, V> {
...
+
+  while (current !== null) {
+    parent = current;
+    if (key < current.key) {
+        current = current.left as Node<K, V>;
+    } else if (key > current.key) {
+        current = current.right as Node<K, V>;
+    } else {
+      // assuming value is an array here
+      (current.value as string[]) = (current.value as string[]).concat(value as string);
```
marco-ippolito, line 252: Array concat is very slow compared to push, might be worth investigating

packages/orama/src/trees/avl.ts
```diff
@@ -198,121 +237,161 @@ export function create<K, V>(key: K, value: V): Node<K, V> {
...
-      node = rotateLeft(node)
-    } else {
-      node.right = rotateRight(node.right as Node<K, V>)
-      node = rotateLeft(node)
+    if (balanceFactor === BALANCE_STATE.UNBALANCED_RIGHT) {
+        if (key > (parent.right as Node<K, V>).key) {
+            parent = rotateLeft(parent);
+        } else {
+            parent.right = rotateRight(parent.right as Node<K, V>);
+            parent = rotateLeft(parent);
```
marco-ippolito, line 286: I guess this block can be simplified since you rotate left in both cases

packages/orama/src/trees/avl.ts
micheleriva: @allevo as discussed privately, we'll open a separate PR if there's anything we can improve

packages/orama/src/trees/avl.ts
```diff
@@ -198,121 +237,161 @@ export function create<K, V>(key: K, value: V): Node<K, V> {
...
-      node.left = rotateLeft(node.left as Node<K, V>)
-      node = rotateRight(node)
+  current = newNode;
+
+  while (parent) {
+    const balanceFactor = getBalanceFactor(parent);
+    
+    if (balanceFactor === BALANCE_STATE.UNBALANCED_LEFT) {
+        if (key < (parent.left as Node<K, V>).key) {
+            parent = rotateRight(parent);
```
marco-ippolito, line 274: I guess this block can be simplified because you rotate right in both cases

packages/orama/src/trees/avl.ts
```diff
@@ -198,121 +237,161 @@ export function create<K, V>(key: K, value: V): Node<K, V> {
...
+        } else {
+            parent.right = rotateRight(parent.right as Node<K, V>);
+            parent = rotateLeft(parent);
+        }
     }
-  }
 
-  return node
-}
+    if (parent == root) {
```
marco-ippolito, line 290: I strongly advise against weak equality, even performance-wise.

packages/orama/src/trees/avl.ts
```diff
@@ -198,121 +237,161 @@ export function create<K, V>(key: K, value: V): Node<K, V> {
...
-  }
+  return root;
+}
 
-  if (key < node.key) {
-    return node.left ? find(node.left, key) : null
-  }
+function getNodeParent<K, V>(root: Node<K, V>, key: K): Node<K, V> | null {
+    let current = root;
+    let parent = null;
```
marco-ippolito, line 303: Same as above no need to initialize

packages/orama/src/trees/avl.ts
```diff
@@ -198,121 +237,161 @@ export function create<K, V>(key: K, value: V): Node<K, V> {
...
+      node = node.right as Node<K, V>;
+    }
   }
 
-  const balanceFactor = getBalanceFactor(node)
-
-  const leftNode = node.left as Node<K, V>
-  const rightNode = node.right as Node<K, V>
+  if (!node) {
+    return null;
```
marco-ippolito, line 350: Usually you want to just return;

packages/orama/src/trees/avl.ts
```diff
@@ -198,121 +237,161 @@ export function create<K, V>(key: K, value: V): Node<K, V> {
...
-      return rotateLeft(node)
+    node.key = minValueNode.key;
+
+    if (minValueParent === node) {
+      minValueParent.right = minValueNode.right;
+    } else {
+      minValueParent.left = minValueNode.right;
     }
+  } else {
+    const childNode = node.left ? node.left : node.right;
```
marco-ippolito, line 381: Use ||= operator

marco-ippolito (commented):
Few nits

micheleriva:
@marco-ippolito could you please send a separate PR?

packages/orama/src/trees/avl.ts
H4ad: There is an excellent article to talk about differences of spread, concat, push, etc... that can bring more ideas: https://jpcamara.com/2023/03/07/making-tanstack-table.html

The general idea is: if `current.value` can mutate, then `.push` is the best option, otherwise, `concat` probably will be faster if you don't use a library like `immer`.

packages/orama/src/trees/avl.ts
H4ad: Removing `Math.abs` is slightly faster:

```
isBalancedOriginal x 3,987,396 ops/sec ±3.63% (87 runs sampled)
isBalancedOptimized x 4,018,579 ops/sec ±2.27% (85 runs sampled)
Fastest is isBalancedOptimized,isBalancedOriginal
```

PR #390, opened 2023-05-26
chore: small code improvements on avl tree

marco-ippolito:
Some code nits


packages/orama/src/trees/avl.ts
```diff
@@ -270,21 +269,19 @@ export function insert<K, V>(root: Node<K, V>, key: K, value: V): Node<K, V> {
...
-        } else {
+        if (key <= (parent.right as Node<K, V>).key) {
             parent.right = rotateRight(parent.right as Node<K, V>);
-            parent = rotateLeft(parent);
         }
+
+        parent = rotateLeft(parent);
     }
 
     if (parent == root) {
```
marco-ippolito, line 287: @micheleriva if we want to check the two object are the same reference we can use `===`.
If we want to check if they have equal value it's a bit more hacky

marco-ippolito, line 287:
```suggestion
    if (parent === root) {
```

packages/orama/src/trees/avl.ts
```diff
@@ -224,32 +224,31 @@ function getNodeByKey<K, V>(node: Node<K, V> | null, key: K): Node<K, V> | null
...
+  while (current !== undefined) {
     parent = current;
     if (key < current.key) {
         current = current.left as Node<K, V>;
     } else if (key > current.key) {
         current = current.right as Node<K, V>;
     } else {
       // assuming value is an array here
-      (current.value as string[]) = (current.value as string[]).concat(value as string);
+      (current.value as string[]).push(...value as string);
```
H4ad, line 251: This will not lead in something like this?

image

micheleriva:
@marco-ippolito not sure why setting undefined over null 👀

marco-ippolito:
> @marco-ippolito not sure why setting undefined over null 👀

There is no reason to initialize a variable with null instead of undefined and also returning null is something quite atypical for javascript, infact you never see it in standard lib. While I do agree that in some cases we might want to set them to null rather than undefined

packages/orama/src/trees/avl.ts
marco-ippolito: You are correct, I need to remoce the spread operator

packages/orama/src/trees/avl.ts
```diff
@@ -351,9 +351,9 @@ export function remove<K, V>(root: Node<K, V> | undefined, key: K): Node<K, V> |
       root = undefined;
     } else {
       if (parentNode.left === node) {
-        parentNode.left = undefined;
+        delete parentNode.left;
```
H4ad, line 354: `delete` is terrible for performance, is better to just set `undefined`.

packages/orama/src/trees/avl.ts
```diff
@@ -224,32 +224,31 @@ function getNodeByKey<K, V>(node: Node<K, V> | null, key: K): Node<K, V> | null
...
+  while (current) {
     parent = current;
     if (key < current.key) {
         current = current.left as Node<K, V>;
     } else if (key > current.key) {
         current = current.right as Node<K, V>;
     } else {
       // assuming value is an array here
-      (current.value as string[]) = (current.value as string[]).concat(value as string);
+      (current.value as string[]).push(...value as string[]);
```
H4ad, line 251: @marco-ippolito If you call insert like this:

```js
const tree = create(1, { foo: 'bar' })

insert(tree, 1, { foo: 'test' });
insert(tree, 1, { foo: 'test2' });
```

It throws an error:

```
file:///home/h4ad/Projects/opensource/performance-analysis/orama-perf-analysis/packages/orama/dist/trees/avl.js:186
            current.value.push(...value);
                                  ^

TypeError: Found non-callable @@iterator
    at insert (file:///home/h4ad/Projects/opensource/performance-analysis/orama-perf-analysis/packages/orama/dist/trees/avl.js:186:27)
    at file:///home/h4ad/Projects/opensource/performance-analysis/orama-perf-analysis/packages/orama/trace-balanced.js:5:1
    at ModuleJob.run (node:internal/modules/esm/module_job:198:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:385:24)
    at async loadESM (node:internal/process/esm_loader:88:5)
    at async handleMainPromise (node:internal/modules/run_main:61:12)
```

H4ad (changes requested):

PR #391, opened 2023-05-26
Changed Wrong Field Name in threshold.mdx

yusufyilmazfr:
Changed database schema field name from title to term.

### DIFF
--- a/packages/docs/pages/usage/search/threshold.mdx
+++ b/packages/docs/pages/usage/search/threshold.mdx
@@ -27,7 +27,7 @@ What happens if I search for `"t-shirt"`?
 
 ```javascript
 const results = await search(db, {
-  title: 't-shirt',
+  term: 't-shirt',
 })
 
 // results.count = 4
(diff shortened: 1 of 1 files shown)


micheleriva:
Hi @yusufyilmazfr ,
nice spot! But I actually made a mistake here:

```
import { create, insert, search } from '@orama/orama'
 
const db = await create({
  schema: {
    title: 'string',
  }
})
 
await insert(db, { term: 'Blue t-shirt, slim fit' })
await insert(db, { term: 'Blue t-shirt, regular fit' })
await insert(db, { term: 'Red t-shirt, slim fit' })
await insert(db, { term: 'Red t-shirt, oversize fit' })
```

we should keep `title` as property and change `term` into `title` when inserting. Could you update the PR that way? Thanks!

yusufyilmazfr:
Hi @micheleriva , I've changed!

micheleriva (approved):
LGMT, thanks!

PR #392, opened 2023-05-26
perf: faster isBalanced AVL Tree method

H4ad:
The current implementation is different than the [old implementation that was recursive](https://github.com/oramasearch/orama/commit/f5e8254ed8203176a17e6d095d91a207060c2c3d#diff-08a9a2514d28f29e668d1ec4cf824b22c5bead9ac9abfb5ed7c6a250f854f8d1L147-L154), actually, the old implementation was faster than the current implementation!

But, I refactored this current implementation to match the old one without the recursion.

```
isBalancedOriginal x 6,168,383 ops/sec ±1.53% (87 runs sampled)
isBalancedOptimizedV4 x 13,253,321 ops/sec ±1.14% (90 runs sampled)
```

<details>
<summary>benchmark.ts</summary>

```ts
import benchmark from 'benchmark';
import { create, insert, Node } from './src/trees/avl.js';

const tree = create(1, { foo: 'bar' })

insert(tree, 2, { foo: 'qux' })
insert(tree, 3, { foo: 'quux' })
insert(tree, 4, { foo: 'quuz' })
insert(tree, 5, { foo: 'corge' })
insert(tree, 6, { foo: 'grault' })
insert(tree, 7, { foo: 'garply' })
insert(tree, 8, { foo: '323' })
insert(tree, 9, { foo: '2311' })

// @ts-ignore
const suite = benchmark.Suite();

console.log(isBalanced(tree));
console.log(isBalancedOptimizedV4(tree));

function getHeight<K, V>(node: Node<K, V> | null): number {
  return node ? node.height : -1
}

type StackNode<K, V> = {
  node: Node<K, V>,
  checkedChildren: boolean
};

export function isBalanced<K, V>(root: Node<K, V> | null): boolean {
  const stack: StackNode<K, V>[] = [];

  if (root !== null) {
    stack.push({ node: root, checkedChildren: false });
  }

  while (stack.length > 0) {
    const top = stack[stack.length - 1];

    if (top.checkedChildren) {
      const heightDiff = Math.abs(getHeight(top.node.left) - getHeight(top.node.right));

      if (heightDiff > 1) {
        return false;
      }

      stack.pop();
    } else {
      top.checkedChildren = true;

      if (top.node.right !== null) {
        stack.push({ node: top.node.right, checkedChildren: false });
      }

      if (top.node.left !== null) {
        stack.push({ node: top.node.left, checkedChildren: false });
      }
    }
  }

  return true;
}

export function isBalancedOptimizedV4<K, V>(root: Node<K, V> | null): boolean {
  if (root === null)
    return true;

  const stack: Node<K, V>[] = [root];

  while (stack.length > 0) {
    const node = stack.pop();

    if (node === undefined)
      return true;

    const heightDiff = getHeight(node.left) - getHeight(node.right);

    if (heightDiff > 1 || heightDiff < -1) {
      return false;
    }

    if (node.right !== null) {
      stack.push(node.right);
    }

    if (node.left !== null) {
      stack.push(node.left);
    }
  }

  return true;
}

suite.add('isBalancedOriginal', () => {
  isBalanced(tree);
});

suite.add('isBalancedOptimizedV4', () => {
  isBalancedOptimizedV4(tree);
});

suite
  // add listeners
  .on('cycle', function (event: any) {
    console.log(String(event.target));
  })
  .on('complete', function () {
    // @ts-ignore
    console.log('Fastest is ' + this.filter('fastest').map('name'));
  })
  .run({
    async: true,
  });
````

</details>

Also, one thing caught my attention, AVL is not always balanced. Why do we need this function?

### DIFF
--- a/packages/orama/src/trees/avl.ts
+++ b/packages/orama/src/trees/avl.ts
@@ -87,33 +87,29 @@ export function getSize<K, V>(root: Node<K, V> | null): number {
 }
 
 export function isBalanced<K, V>(root: Node<K, V> | null): boolean {
-  const stack: StackNode<K, V>[] = [];
+  if (root === null)
+    return true;
 
-  if (root !== null) {
-    stack.push({ node: root, checkedChildren: false });
-  }
+  const stack: Node<K, V>[] = [root];
 
   while (stack.length > 0) {
-    const top = stack[stack.length - 1];
+    const node = stack.pop();
 
-    if (top.checkedChildren) {
-      const heightDiff = Math.abs(getHeight(top.node.left) - getHeight(top.node.right));
+    if (node === undefined)
+      return true;
 
-      if (heightDiff > 1) {
-        return false;
-      }
+    const heightDiff = getHeight(node.left) - getHeight(node.right);
 
-      stack.pop();
-    } else {
-      top.checkedChildren = true;
+    if (heightDiff > 1 || heightDiff < -1) {
+      return false;
+    }
 
-      if (top.node.right !== null) {
-        stack.push({ node: top.node.right, checkedChildren: false });
-      }
+    if (node.right !== null) {
+      stack.push(node.right);
+    }
 
-      if (top.node.left !== null) {
-        stack.push({ node: top.node.left, checkedChildren: false });
-      }
+    if (node.left !== null) {
+      stack.push(node.left);
     }
   }
 


micheleriva (approved):
That's fantastic, thanks!

PR #394, opened 2023-05-28
perf: skip get balance factor

H4ad:
Since `getBalanceFactor` is used in one place, don't have the necessity of having a function with switch statement, just get the diff of the height and then compare directly.

I also removed the unused StackNode type.


micheleriva (approved):
makes sense, thanks! lgtm

PR #398, opened 2023-05-29
Feature/group by api

allevo:
This PR introduces two new capabilities:
- group the result by one or more properties `groupBy`
- make a uniqueness of a value in a property `distinctOn`

Suppose your DB is something like:
```typescript
await insertMultiple(db, [
  { id: '0', type: 't-shirt', design: 'A', color: 'blue', rank: 3, isPromoted: true },
  { id: '1', type: 't-shirt', design: 'A', color: 'green', rank: 5, isPromoted: false },
  { id: '2', type: 't-shirt', design: 'A', color: 'red', rank: 4, isPromoted: false },
  { id: '3', type: 't-shirt', design: 'B', color: 'blue', rank: 4, isPromoted: false },
  { id: '4', type: 't-shirt', design: 'B', color: 'green', rank: 4, isPromoted: true },
  { id: '5', type: 't-shirt', design: 'B', color: 'white', rank: 5, isPromoted: false },
  { id: '6', type: 't-shirt', design: 'B', color: 'gray', rank: 5, isPromoted: true },
  { id: '7', type: 'sweatshirt', design: 'A', color: 'yellow', rank: 3, isPromoted: true },
  { id: '8', type: 'sweatshirt', design: 'A', color: 'green', rank: 4, isPromoted: false },
])
```
Every document is a variation of an item.

# `groupBy`

Orama supports the following query to group all the variants per design value keeping only the first element sorted by rank.
```typescript
  term: 't-shirt',
  groupBy: {
    properties: ['design'], // property on which we want to group on
    maxResult: 1, // for every group, how many results we want
  },
  sortBy: {
    property: 'rank', // inside a group, the result is ordered following this property
    order: 'DESC', // with this order
  },
})
```
Example of the returned object
```
{
  groups: [
    {
      values: ['A'],
      result: [
        {
          id: '1',
          score: 0, 
          document: { id: '1', type: 't-shirt', design: 'A', color: 'green', rank: 5, isPromoted: false } 
        }
      ]
    },
    {
      values: ['B'],
      result: [
        {
          id: '5',
          score: 0, 
          document: { id: '5', type: 't-shirt', design: 'B', color: 'white', rank: 5, isPromoted: false }
        }
      ]
    }
  ],
  // The other common properties like `hits` and `elapsed`
}
```

Multiple properties are supported as well
```typescript
  groupBy: {
    properties: ['design', 'rank', 'isPromoted'], // group on the combination of the values
  },
  sortBy: {
    property: 'id',
    order: 'ASC',
  },
})
```

A custom reducer can be specified as well
```typescript
interface Doc extends Document {
  type: string
  design: string
  rank: number
  color: string
  isPromoted: boolean
}
interface AggregationValue {
  type: string
  design: string
  colors: string[]
  ranks: number[]
  isPromoted: boolean
}

  groupBy: {
    properties: ['type', 'design'],
    reduce: {
      reducer: (_: ScalarSearchableValue[], acc: AggregationValue, item: Result) => {
        const doc = item.document as Doc
        acc.type ||= doc.type
        acc.design ||= doc.design
        acc.isPromoted ||= doc.isPromoted
        acc.colors.push(doc.color)
        acc.ranks.push(doc.rank)
        return acc
      },
      getInitialValue: (): AggregationValue => ({ type: '', design: '', colors: [], ranks: [], isPromoted: false }),
    },
  },
  sortBy: {
    property: 'rank',
    order: 'DESC',
  }
})
```
The above code allows you to group all the variants into a unique document.

NB: the groups themselves are not sorted

# `distinctOn`
This PR also implements the `distinctOn` capability: the `hits` array is filtered by the uniqueness of a key. Let's consider the DB above. We can implement the `distinct` functionality below
```typescript
  distinctOn: 'type',
  sortBy: {
    property: 'rank',
    order: 'DESC',
  }
})
```
The returned value is like
```
{
  hits: [
    { id: 1, score: 0, document: { id: '1', type: 't-shirt', design: 'A', color: 'green', rank: 5, isPromoted: false } },
    { id: 5, score: 0, document: { id: '5', type: 't-shirt', design: 'B', color: 'white', rank: 5, isPromoted: false } }
  ],
  count: 100, 
}
```
This allows returning documents under `hits` uniquely on the `distinctOn` property: only the first document with a new value will be returned, and the following ones will be skipped.
This other proposal doesn't allow us to calculate an aggregation on the distinct group and, for this reason, should be faster than the one proposed in the PR.
The count property is the total document count that matches the query ignoring the uniqueness.

# TODO:
- documentation
- ~~values should be with the right types~~

### DIFF
--- a/packages/orama/src/components/groups.ts
+++ b/packages/orama/src/components/groups.ts
@@ -1,0 +1,168 @@
+import type { Orama, ScalarSearchableValue, TokenScore, GroupByParams, GroupResult, Result } from '../types.js'
+import { getNested, intersect } from '../utils.js'
+
+interface PropertyGroup {
+  property: string
+  perValue: Record<
+    string,
+    {
+      indexes: number[]
+      count: number
+    }
+  >
+}
+
+export async function getGroups(orama: Orama, results: TokenScore[], by: GroupByParams): Promise<GroupResult> {
+  const groupBy = by!
+  const allIDs = results.map(([id]) => id)
+
+  // allDocs is already sorted by the sortBy algorithm
+  // We leverage on that to limit the number of documents returned
+  const allDocs = await orama.documentsStore.getMultiple(orama.data.docs, allIDs)
+  const allDocsLength = allDocs.length
+
+  const returnedCount = groupBy.maxResult || Number.MAX_SAFE_INTEGER
+  const properties = groupBy.property ? [groupBy.property] : groupBy.properties || []
+  const propertiesLength = properties.length
+
+  const values: string[][] = []
+
+  // We want to understand which documents have which values
+  // and group them by the property and values
+  const g: Record<string, PropertyGroup> = {}
+  for (let i = 0; i < propertiesLength; i++) {
+    const groupByKey = properties[i]
+    const group: PropertyGroup = {
+      property: groupByKey,
+      perValue: {},
+    }
+
+    for (let j = 0; j < allDocsLength; j++) {
+      const doc = allDocs[j]
+
+      let value = await getNested<ScalarSearchableValue>(doc as object, groupByKey)
+      // TODO: should we consider undefined values?
+      if (typeof value === 'undefined') {
+        continue
+      }
+      if (typeof value === 'boolean') {
+        value = value ? 'true' : 'false'
+      }
+      if (typeof group.perValue[value] === 'undefined') {
+        group.perValue[value] = {
+          indexes: [],
+          count: 0,
+        }
+      }
+      if (group.perValue[value].count >= returnedCount) {
+        // We stop early because for this value we react the limit
+        continue
+      }
+
+      // We use the index to keep track of the original order
+      group.perValue[value].indexes.push(j)
+      group.perValue[value].count++
+    }
+
+    values.push(Object.keys(group.perValue))
+
+    g[groupByKey] = group
+  }
+
+  const combinations = calculateCombination(values)
+  const combinationsLength = combinations.length
+
+  const groups: {
+    values: string[]
+    indexes: number[]
+  }[] = []
+  for (let i = 0; i < combinationsLength; i++) {
+    const combination = combinations[i]
+    const combinationLength = combination.length
+
+    const group: {
+      values: string[]
+      indexes: number[]
+    } = {
+      values: [],
+      indexes: [],
+    }
+    const indexes: number[][] = []
+    for (let j = 0; j < combinationLength; j++) {
+      const value = combination[j]
+      const property = properties[j]
+      indexes.push(g[property].perValue[value].indexes)
+      group.values.push(value)
+    }
+    // We leverage on the index to sort the results by the original order
+    group.indexes = intersect(indexes).sort((a, b) => a - b)
+
+    // don't generate empty groups
+    if (group.indexes.length === 0) {
+      continue
+    }
+
+    groups.push(group)
+  }
+
+  const res: GroupResult = []
+
+  const groupsLength = groups.length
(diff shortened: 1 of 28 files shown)


packages/orama/src/components/groups.ts
```diff
@@ -0,0 +1,171 @@
...
+    res.push({
+      values: group.values,
+      result: aggregationValue,
+    })
+  }
+
+  return res
+}
+
+function calculateCombination(arrs: ScalarSearchableValue[][]): ScalarSearchableValue[][] {
```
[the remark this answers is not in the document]
micheleriva: We could even use generators there, something like:

```js
function* generateCombinations(arrs, currentCombination = [], index = 0) {
  if(index < arrs.length) {
    for(let val of arrs[index]) {
      yield* generateCombinations(arrs, [...currentCombination, val], index + 1);
    }
  } else {
    yield currentCombination;
  }
}

function calculateCombination(arrs) {
  return [...generateCombinations(arrs)];
}
```

but maybe I'm totally wrong so rest in pepperoni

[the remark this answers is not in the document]
allevo: I added some benchmarks.
Substitute the implementation with the `calculateCombination_Recursive_Head_Indexed` that improves the ops/sec by 40% and more.

packages/orama/src/components/sorter.ts
```diff
@@ -170,35 +168,23 @@ async function sortBy(sorter: Sorter, docIds: [string, number][], by: SorterPara
...
-  let unsortableDocumentTotal = 0
-  if (isDesc) {
-    for (let i = 0; i < docIdsLength; i++) {
-      if (typeof s.docs[docIds[i][0]] === 'undefined') {
-        unsortableDocumentTotal++
-      }
+  docIds.sort((a, b) => {
+    const posA = s.docs[a[0]]
+    const posB = s.docs[b[0]]
+    if (typeof posA === 'undefined' && typeof posB === 'undefined') {
```
micheleriva, line 174: We're duplicating the `typeof` operation for both `posA` and `posB`. I'd move them in a single, separate constant

packages/orama/src/components/sorter.ts
```diff
@@ -170,35 +168,23 @@ async function sortBy(sorter: Sorter, docIds: [string, number][], by: SorterPara
...
-  // Calculate how many documents aren't inside the sorter index.
-  // Used only for "DESC" sort.
-  let unsortableDocumentTotal = 0
-  if (isDesc) {
-    for (let i = 0; i < docIdsLength; i++) {
-      if (typeof s.docs[docIds[i][0]] === 'undefined') {
-        unsortableDocumentTotal++
-      }
+  docIds.sort((a, b) => {
+    const posA = s.docs[a[0]]
```
micheleriva, line 172: I'd give a more mnemonic name to `posA` and `posB` variables

packages/orama/src/types.ts
```diff
@@ -273,6 +290,17 @@ export type FacetResult = Record<
...
 
+export type GroupResult =
+  | {
+      values: ScalarSearchableValue[]
+      result: Result[]
+    }[]
+  | {
+      values: ScalarSearchableValue[]
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      result: any
```
micheleriva, line 301: No way to remove this `any`, right?

micheleriva (approved):
Mostly ok, just code style indications. Looks good to me

packages/orama/src/types.ts
allevo: resolved forwarding the type from SearchParams to Results.

PR #403, opened 2023-05-30
changed array item order in sorting.mdx

yusufyilmazfr:


packages/docs/pages/usage/search/sorting.mdx
```diff
@@ -108,9 +108,10 @@ const db = await create({
 ```
 
 The function accepts 2 parameter with the following format `[string, number, Document]` that stands for:
-1. id of the document
-2. score of the document
-3. the document
+
+0. id of the document
```
yusufyilmazfr, line 112: It's not important but a little confusing. We'are talking about array but array parameters start with 1.

micheleriva: @allevo what do you think?

allevo: I understand the point but never saw a list starting from 0.
We can resolve this point by putting a bulleted list instead of a numbered one.
What do you think?

micheleriva: @yusufyilmazfr let's go for a bulleted list

yusufyilmazfr: @micheleriva done! 🤞

micheleriva (approved):
lgtm

PR #404, opened 2023-05-31
feat: reworked docusaurus plugin to support docs versions.

ShogunPanda:
This PR completely reworks the Docusaurus plugin to be more lean and to support versioning.


packages/plugin-docusaurus/src/client/theme/SearchBar/index.tsx
```diff
@@ -64,42 +65,64 @@ function snippet(item: Hit): JSX.Element {
...
-
+      async getSources({ query: term }): Promise<any> {
         return [
           {
             sourceId: 'orama',
             async getItems() {
-              const results = await orama(query)
+              const results = await searchWithHighlight(database, {
+                term,
+                properties: ['sectionTitle', 'sectionContent', 'type']
```
allevo, line 110: Would you mind searching only on these properties? Should we allow customization?

packages/plugin-docusaurus/src/client/theme/SearchBar/index.tsx
```diff
@@ -109,7 +132,16 @@ export default function SearchBar(): JSX.Element {
...
-            <Footer />
+            <Footer colorMode={colorMode} />
+          </>,
+          root
+        )
+      },
+      renderNoResults({ render, state }, root) {
+        render(
+          <>
+            {state.query && <div className="aa-NoResults">No results found.</div>}
```
allevo, line 143: Not sure `state.query` is the right value that we should leverage. Is it correct?

packages/plugin-docusaurus/src/client/theme/SearchBar/index.tsx
```diff
@@ -118,7 +150,42 @@ export default function SearchBar(): JSX.Element {
...
+
+  useEffect(() => {
+    async function loadDatabase(): Promise<void> {
+      let data: RawDataWithPositions
+      if (!searchData) {
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        const searchResponse = await fetch(searchBaseUrl.replace('@VERSION@', version!))
+
+        if (searchResponse.status === 0) {
+          throw new Error(`Network error: ${await searchResponse.text()}`)
```
allevo, line 163: throwing errors in `useEffect` can cause very bad behavior. Should we put a generic error message for say "OOOPS something goes wrong" to the user?

packages/plugin-docusaurus/src/client/theme/SearchBar/index.tsx
```diff
@@ -118,7 +150,42 @@ export default function SearchBar(): JSX.Element {
...
+        } else if (searchResponse.status !== 200) {
+          throw new Error(`HTTP error ${searchResponse.status}: ${await searchResponse.text()}`)
+        }
+
+        const deflated = ungzip(await searchResponse.arrayBuffer(), { to: 'string' })
```
allevo, line 168: why we should to this by ourselves? if the content encoding is right, the fetch should to this automatically.

packages/plugin-docusaurus/src/client/theme/SearchBar/index.tsx
ShogunPanda: Right now the properties come from the parse-doc plugin, so they are fixed. We can eventually revisit in the future.

packages/plugin-docusaurus/src/client/theme/SearchBar/index.tsx
ShogunPanda: AFAIK I got from the widget documentation yes.

packages/plugin-docusaurus/src/client/theme/SearchBar/index.tsx
ShogunPanda: I had exactly your doubt.
I chose to manually compress to make sure this is always handled correctly so that users get small JSON to be loaded. For instance, in the fastify website the uncompressed data is 8MB while the compressed one is around 1MB.

packages/plugin-astro/package.json
```diff
@@ -36,7 +36,7 @@
     "dev": "swc --delete-dir-on-start -s -w --extensions .ts -d dist src",
     "build": "swc --delete-dir-on-start --extensions .ts -d dist src",
     "postbuild": "tsc -p . --emitDeclarationOnly",
-    "-test": "node --loader=tsx --no-warnings=loader --test ./test/integration.ts | tap-mocha-reporter spec",
+    "test:disabled": "node --loader=tsx --no-warnings=loader --test ./test/integration.ts | tap-mocha-reporter spec",
```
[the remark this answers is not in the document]
ShogunPanda: Yup, I can. I'm not sure when they got disabled but I will revert this.

packages/plugin-docusaurus/package.json
```diff
@@ -33,7 +33,7 @@
     "prebuild": "npm run lint",
     "build": "swc --delete-dir-on-start --extensions .ts,.tsx,.cts -d dist src",
     "postbuild": "sh scripts/postbuild.sh",
-    "-test": "node --loader=tsx --no-warnings=loader --test ./test/integration.ts | tap-mocha-reporter spec",
+    "test:disabled": "node --loader=tsx --no-warnings=loader --test ./test/integration.ts | tap-mocha-reporter spec",
```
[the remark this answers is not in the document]
ShogunPanda: Yup, I can. I'm not sure when they got disabled but I will revert this.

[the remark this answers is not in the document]
ShogunPanda: Done!

packages/plugin-astro/package.json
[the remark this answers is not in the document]
ShogunPanda: Done!

micheleriva:
@ShogunPanda  there's still one last error on the Vercel side:

Screenshot 2023-06-02 at 08 48 47

Issue #405, opened 2023-05-31
Orama Astro | Plugin | Unable to build on Windows


TheZalRevolt:
**Describe the bug**
When the build command is executed on Windows the build fail due to an error on the destination path of the index.html file.

**To Reproduce**
1. Create a new Astro Project
2. Install @orama/plugin-astro
3. Setup the plugin
4. Run astro build

**Expected behavior**
The build to complete

**Screenshots**
![build error](https://github.com/oramasearch/orama/assets/129967922/77d3d312-d1f8-4d22-bb99-15d4fbc17ce0)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Version: @orama/plugin-astro@1.0.2

![version](https://github.com/oramasearch/orama/assets/129967922/86e84202-220a-4294-92a1-4901b17e342f)

micheleriva:
@Balastrong any idea? 👀

micheleriva:
Fixed in #406

PR #406, opened 2023-05-31
fix: fix the build of Astro Plugin on Windows OS

TheZalRevolt:
After trying to run a build of my Astro Blog with the Orama Plugin I noticed an error.
It's because some directories aren't compliant with Windows Os.

I also fixed the fact that the plugin was looking for files with wrong names like `mobile-appindex.html` instead of `mobile-app/index.html`

After the fix the build complete without any other issue.

This closes issue #405

### DIFF
--- a/packages/plugin-astro/src/index.ts
+++ b/packages/plugin-astro/src/index.ts
@@ -65,16 +65,16 @@ async function prepareOramaDb(
   })
 
   // All routes are in the same folder, we can use the first one to get the basePath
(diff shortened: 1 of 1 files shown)


TheZalRevolt:
Thank you very much @Balastrong !

Balastrong (approved):
I still wonder how other Astro plugins are currently handling that weird Windows behavior of adding a leading `/` on paths...

Anyway, so far so good, LGTM :)

micheleriva (approved):
lgtm! Thanks!

PR #411, opened 2023-06-06
fix: Fixed e2e tests.

ShogunPanda:
This fixes E2E for Astro and Docusaurus when cutting a new version. 
It also fixes a new problem in MDX.

### DIFF
--- a/packages/plugin-astro/test/integration.ts
+++ b/packages/plugin-astro/test/integration.ts
@@ -2,7 +2,7 @@ import { create as createOramaDB, load as loadOramaDB, Schema, search } from '@o
 import assert from 'node:assert'
 import { exec, ExecException } from 'node:child_process'
 import { existsSync } from 'node:fs'
-import { cp, readFile, rm } from 'node:fs/promises'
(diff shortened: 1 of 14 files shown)


packages/plugin-astro/test/integration.ts
```diff
@@ -17,7 +17,7 @@ interface Execution {
 }
 
 const sandboxSource = fileURLToPath(new URL('./sandbox', import.meta.url))
-const sandbox = resolve(tmpdir(), `orama-plugin-astro-${Date.now()}`)
+const sandbox = process.env.KEEP_SANDBOX ? '/tmp/orama-astro-sandbox' : resolve(tmpdir(), `orama-plugin-astro-${Date.now()}`)
```
allevo, line 20: Will `'/tmp/orama-astro-sandbox'` work on Windows?

ShogunPanda: Good point. It will not. I'll send a little small PR for this. Anyway it's just a debugging flag mostly for the core team so it's not really important.

PR #430, opened 2023-07-01
feat(orama): adds zip tree

micheleriva:
This PR aims to add an initial implementation of a Zip Tree to substitute AVL Trees on indexes with a lot of numeric properties. It's 100% backward compatible with the AVL Trees and exposes the same APIs.


hastebrot:
Looks great!

I'm not sure, `Math.random()` should be replaced by `randomRank()`, but I have to look more closer.

```ts
function randomRank(): number {
  let heads = 0;
  while (Math.random() < 0.5) {
    heads += 1;
  }
  return heads;
}
```

micheleriva:
> > @hastebrot thanks for the suggestion, I added it as follows:
> > ```ts
> > function randomRank (maxAttempts = 32): number {
> > 
> >   let heads = 0
> > 
> > 
> > 
> >   for (let i = 0; i < maxAttempts; i++) {
> > 
> >     if (Math.random() >= 0.5) break
> > 
> >     heads += 1
> > 
> >   }
> > 
> > 
> > 
> >   return heads
> > 
> > }
> > ```
> 
> I was thinking if adding a better random algorithm like one from stdlibjs?

It's ok, but we can't use external dependencies.

hastebrot:
```js
// Function to generate a random number from a geometric distribution.
function randomRank(p, r) {
    return Math.floor(Math.log(1 - r) / Math.log(1 - p)) + 1;
}

const p = 0.5; // Probability of getting a head in a fair coin flip.
const r = Math.random(); // Generate a random number between 0 and 1.
const numberOfFlips = geometricRandom(p, r);
```

hastebrot:
@thebergamo it's a jungle of code but I managed to find the line with geometric distribution in stdlibjs. what is cool is that they even have a proof for it in the jsdocs.

https://github.com/stdlib-js/stdlib/blob/v0.0.96/lib/node_modules/%40stdlib/random/base/geometric/lib/geometric.js#L78

but in the end it's really just a single line of code.

micheleriva:
I'm going to merge this PR without replacing AVL tree for the number index. I'll dedicate a separate PR for this. Thanks @hastebrot for the great help 🙏

PR #431, opened 2023-07-02
perf: use binary search instead of linear search

H4ad:
I notice the massive slowdown on the insert due to the `stringSort`, since the array is ordered, we can use `binarySearch` instead of using `linearSearch` to add new items.

For the following code went from `1m 7s` to `30s` with this change:

```js
import { create, insert } from "./dist/index.js";

(async () => {
  const db = await create({
    schema: {
      name: "string"
    }
  });

  for (let i = 0; i < 3e4; i++) {
    await insert(db, {
      id: Math.random().toString(16).slice(2).repeat(3),
    });
  }

  console.log("Memory usage: ", process.memoryUsage().heapTotal / 1024 / 1024, "MB");
})();
```

The only problem with this current implementation is: we can't guarantee the insertion order when we add the same value multiple times, and that broke the `group.spec.ts` test, so what do you want to do in this case? We can add a special case to guarantee the same behavior, but I don't think it's worth the operations spent just to guarantee the order.


micheleriva:
That's fantastic. Could you please fix the failing tests?

H4ad:
Just a note, with this change we don't guarantee that repeated values will return the same value, as in this test: https://github.com/oramasearch/orama/pull/431/files#diff-7a7a4f13f68b36b3011a9c2de13afd3f7f59954cdb421984ac690fe4ae2a7767R321

If any value is added to `orderedDocs`, a different value may be returned, as it returns the first matching value.

micheleriva:
@allevo wdyt?

H4ad:
Actually, I think I have a better way to fix the insert slowdown without using `binarySearch`.

Give me some time to test before merging this PR.

With the new way, this PR became useless.

micheleriva:
@H4ad should we close this?

H4ad:
@H4ad not yet, let me test first my idea, I will open another PR if the new idea works or I came back with the result of my failed exploration.

H4ad:
@micheleriva Continued at #434

PR #434, opened 2023-07-03
perf: lazy operations on sorter

### LINKED ISSUE #348: Abnormal memory use
Hello everyone,
I am trying _Orama_ to do research on a **dataset of 355000 documents**. 

Each document has the following pattern: 
```
{
   "id":"f550afbe-8a3f-4ec5-9bff-19590869dfd0", // v4 uuid string
   "name":"Foo",  // string
   "surname":"Bar", // string
   "fiscalCode":"NTNPFR53E06B058O", // italian codice fiscale as string
   "season":"2223" // 4 integer as string
}
```
I should do the text search on `nameSurname` and `fiscalCode`, but I need the other fields as filters.

- How much memory should the search database occupy?
- On k8s with the 2000Mi limit (2gb) it would seem not to be enough. Is this normal?

In any case, I like Orama very much, you have done a super job. Thank you!

H4ad:
Well, instead of sorting every time you insert an item, now it only sort when you need to search, this speedup up the insert by A LOT.

To compare, the following code went from `1m 7s` to `1s` with this change:

```js
import { create, insert, remove, search } from "./dist/index.js";

(async () => {
  const db = await create({
    schema: {
      name: "string"
    }
  });

  let now = performance.now();
  for (let i = 0; i < 3e4; i++) {
    await insert(db, {
      id: i.toString(),
    });
  }
  console.log(`insert: ${performance.now() - now}ms`);
  console.log(`memory: ${process.memoryUsage().heapUsed / 1024 / 1024}MB`);
  now = performance.now();

  await search(db, {
    term: 'lll'
  });
  console.log(`search: ${performance.now() - now}ms`);
  now = performance.now();

  for (let i = 0; i < 1e4; i++) {
    await remove(db, i.toString());
  }
  console.log(`remove: ${performance.now() - now}ms`);
  console.log(`memory: ${process.memoryUsage().heapUsed / 1024 / 1024}MB`);
  now = performance.now();

  await search(db, {
    term: 'lll'
  });
  console.log(`search: ${performance.now() - now}ms`);
})();
```

Before this change:

```
insert: 49947.061446000356ms
memory: 44.56006622314453MB
search: 1.312158000189811ms
remove: 5246.693398999982ms
memory: 45.88232421875MB
search: 0.10874100029468536ms
```

Now:

```
insert: 694.5329680000432ms
memory: 54.184051513671875MB
search: 1.2686470001935959ms
remove: 92.8359210002236ms
memory: 58.10900115966797MB
search: 0.10343399969860911ms
```

---

Also, this new change also fix the #348, instead of hours, it took just a minute to run.

/claim #434

### DIFF
--- a/packages/orama/src/components/sorter.ts
+++ b/packages/orama/src/components/sorter.ts
@@ -1,5 +1,5 @@
-import { createError } from '../errors.js';
-import { ISorter, OpaqueSorter, Orama, Schema, SorterConfig, SorterParams, SortType, SortValue } from '../types.js';
+import { createError } from '../errors.js'
+import { ISorter, OpaqueSorter, Orama, Schema, SorterConfig, SorterParams, SortType, SortValue } from '../types.js'
 
 interface PropertySort<K> {
   docs: Map<string, number>
@@ -120,7 +120,8 @@ function ensureIsSorted(sorter: Sorter): void {
     return
   }
 
-  for (const prop of Object.keys(sorter.sorts)) {
+  const properties = Object.keys(sorter.sorts)
+  for (const prop of properties) {
     ensurePropertyIsSorted(sorter, prop);
   }
 
@@ -166,7 +167,8 @@ function ensurePropertyIsSorted(sorter: Sorter, prop: string): void {
 }
 
 function ensureOrderedDocsAreDeleted(sorter: Sorter): void {
-  for (const prop of Object.keys(sorter.sorts)) {
+  const properties = Object.keys(sorter.sorts)
+  for (const prop of properties) {
     ensureOrderedDocsAreDeletedByProperty(sorter, prop)
   }
 }
--- a/packages/orama/src/components/sorter.ts
+++ b/packages/orama/src/components/sorter.ts
@@ -1,15 +1,15 @@
-import { createError } from "../errors.js";
-import { ISorter, OpaqueSorter, Orama, Schema, SorterConfig, SorterParams, SortType, SortValue } from "../types.js";
+import { createError } from '../errors.js';
+import { ISorter, OpaqueSorter, Orama, Schema, SorterConfig, SorterParams, SortType, SortValue } from '../types.js';
 
 interface PropertySort<K> {
   docs: Record<string, number>
   orderedDocs: [string, K][]
-  type: SortType;
+  type: SortType
 }
 
 export interface Sorter extends OpaqueSorter {
-  language?: string;
-  isSorted: boolean;
+  language?: string
+  isSorted: boolean
   enabled: boolean
   sortableProperties: string[]
   sortablePropertiesWithTypes: Record<string, SortType>
@@ -261,6 +261,8 @@ export async function save<R = unknown>(sorter: Sorter): Promise<R> {
       enabled: false,
     } as unknown as R
   }
+
+  ensureIsSorted(sorter)
 
   return {
     sortableProperties: sorter.sortableProperties,
--- a/packages/orama/src/components/sorter.ts
+++ b/packages/orama/src/components/sorter.ts
@@ -2,10 +2,13 @@ import { createError } from '../errors.js';
 import { ISorter, OpaqueSorter, Orama, Schema, SorterConfig, SorterParams, SortType, SortValue } from '../types.js';
 
 interface PropertySort<K> {
-  docs: Record<string, number>
+  docs: Map<string, number>
   orderedDocs: [string, K][]
+  orderedDocsToRemove: Map<string, boolean>
   type: SortType
 }
+
+type SerializablePropertySort<K> = Omit<PropertySort<K>, 'orderedDocsToRemove' | 'docs'> & { docs: Record<string, number> } ;
 
 export interface Sorter extends OpaqueSorter {
   language?: string
@@ -58,7 +61,8 @@ function innerCreate(schema: Schema, sortableDeniedProperties: string[], prefix:
         sorter.sortableProperties.push(path)
         sorter.sortablePropertiesWithTypes[path] = type
         sorter.sorts[path] = {
-          docs: {},
+          docs: new Map(),
+          orderedDocsToRemove: new Map(),
           orderedDocs: [],
           type: type,
         }
@@ -103,7 +107,7 @@ async function insert(
 
   const s = sorter.sorts[prop]
 
-  s.docs[id] = s.orderedDocs.length;
+  s.docs.set(id, s.orderedDocs.length);
   s.orderedDocs.push([id, value]);
 }
 
@@ -157,8 +161,24 @@ function ensurePropertyIsSorted(sorter: Sorter, prop: string): void {
   const orderedDocsLength = s.orderedDocs.length;
   for (let i = 0; i < orderedDocsLength; i++) {
     const docId = s.orderedDocs[i][0]
-    s.docs[docId] = i;
+    s.docs.set(docId, i);
   }
+}
+
+function ensureOrderedDocsAreDeleted(sorter: Sorter): void {
+  for (const prop of Object.keys(sorter.sorts)) {
+    ensureOrderedDocsAreDeletedByProperty(sorter, prop)
+  }
+}
+
+function ensureOrderedDocsAreDeletedByProperty(sorter: Sorter, prop: string): void {
+  const s = sorter.sorts[prop]
+
+  if (!s.orderedDocsToRemove.size)
+    return
+
+  s.orderedDocs = s.orderedDocs.filter(doc => !s.orderedDocsToRemove.has(doc[0]))
+  s.orderedDocsToRemove.clear()
 }
 
 async function remove(sorter: Sorter, prop: string, id: string) {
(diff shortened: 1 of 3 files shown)


micheleriva:
@H4ad how does this affect search time?

H4ad:
image

Just add 40ms on the first search.

You can also call `ensureIsSorted` after insert to not affect the `search`.

Now it probably consumes 99% less time, is it worth a bounty? e.e

micheleriva:
> You can also call ensureIsSorted after insert to not affect the search.

Could you please add this as a last function call in `insertMultiple`?

 ---

@allevo maybe we should expose a simple `sort` function to ensure sorting? Something like:

```js
import { create, insertMultiple, ensureSort } from '@orama/orama'

const db = await create({ ... })

await insertMultiple(db, [...])

await ensureSort(db)
```

^ super bad DX, written it down not to forget about it

H4ad:
@micheleriva Only worth adding if you know the users never call `insertMultiple` twice, do you still want to add it?

ShogunPanda:
The PR looks fine to me in general, even though I don't really like the `lazy` approach to solve the performance problem.
I'm conflicted on where to put `ensureSort`.

As @H4ad pointed out, doing it within `insertMultiple` might make not a lot of sense if is not called iteratively.

If we move the `.sorted` check from `ensurePropertyIsSorted` to `ensureIsSorted` it should probably be fine to call within search.

micheleriva:
What about this:

1. Add a `isSorted` boolean property to the Orama instance
2. Running a function such as `ensureSort` will set it to `true`
3. If `isSorted` is set to `false`, the first search operation will call `ensureSort`
4. After every new insertion/update, `isSorted` should route back to `false`

what do you think?

ShogunPanda:
So we handle this outside of the sorter? Ok, it makes sense to me.

@H4ad Can you please implement the aforementioned changes?

micheleriva:
@H4ad we didn't plan this activity, but we'll reward a small bounty anyway as soon as it lands. Your work is immensely appreciated

H4ad:
As soon as I have time I will implement these changes, probably today.

H4ad:
@micheleriva I read the suggestions but it doesn't make sense to have the `isSorted` inside the `orama` instance.

If I define the `isSorted` on `orama` instance, I don't know which property is not sorted since we keep the `orderedDocs` inside an `Record` of properties.

Or, I keep the current changes and I also add a property called `isSorted`, to when I call `insert` I mark the `PropertySort` and also the `orama` instance as `isSorted=false`, what do you think?

ShogunPanda:
@H4ad The idea is that you don't really remember which properties are sorted and which are not. 

It's easier to just assume that when a document is inserted/updated/removed all properties are unsorted. The rationale behind this is that unliked other system Orama is optimized to be used in a two phases: "insert all documents" and then "search".
All heavy duties are done in the inserting phase while the searching phase must be the most lightweight possible.

H4ad:
@ShogunPanda I agree with your design of Orama and this change will not affect that.

Calling `ensureSort` was just used by the tests, it will not be used by the users because they don't need it.

Doing this approach will just make the `search` slower since we need to traverse NItems*NProperties, instead of traversing the properties that were modified.

On the optimization side, adding it to `orama` does not make sense, but if you still want to do that, I will push a commit later this day.

ShogunPanda:
If you don't want to add to Orama, you can add it to the sorted, eventually.

H4ad:
@ShogunPanda Added isSorter to `sorter` and also I added the properties to be serialized, so we can restore the sorter without needing to sort again.

H4ad:
I also transform the remove operation into lazy one, before:

```
insert: 49947.061446000356ms
memory: 44.56006622314453MB
search: 1.312158000189811ms
remove: 5246.693398999982ms
memory: 45.88232421875MB
search: 0.10874100029468536ms
```

Now:

```
insert: 694.5329680000432ms
memory: 54.184051513671875MB
search: 1.2686470001935959ms
remove: 92.8359210002236ms
memory: 58.10900115966797MB
search: 0.10343399969860911ms
```

packages/orama/src/components/sorter.ts
```diff
@@ -54,7 +61,8 @@ function innerCreate(schema: Schema, sortableDeniedProperties: string[], prefix:
         sorter.sortableProperties.push(path)
         sorter.sortablePropertiesWithTypes[path] = type
         sorter.sorts[path] = {
-          docs: {},
+          docs: new Map(),
```
micheleriva, line 64: I fear we can't use maps here as they're not directly serializable to JSON. @allevo, @ShogunPanda please correct me if I'm wrong

packages/orama/src/components/sorter.ts
```diff
@@ -54,7 +61,8 @@ function innerCreate(schema: Schema, sortableDeniedProperties: string[], prefix:
...
+          orderedDocsToRemove: new Map(),
```
micheleriva, line 65: We can use a map here as long as we don't need to serialize this property to JSON. As far as I understand, we can leave this property as Map.

packages/orama/src/components/sorter.ts
```diff
@@ -103,56 +101,99 @@ async function insert(
...
+function ensureIsSorted(sorter: Sorter): void {
+  if (sorter.isSorted) {
+    return
+  }
+
+  if (!sorter.enabled) {
+    return
+  }
+
+  for (const prop of Object.keys(sorter.sorts)) {
```
micheleriva, line 123: Move `Object.keys` outside the loop condition to avoid accessing the object keys on every iteration

packages/orama/src/components/sorter.ts
```diff
@@ -103,56 +101,99 @@ async function insert(
...
+  const orderedDocsLength = s.orderedDocs.length;
+  for (let i = 0; i < orderedDocsLength; i++) {
     const docId = s.orderedDocs[i][0]
-    s.docs[docId]++
+    s.docs.set(docId, i);
+  }
+}
+
+function ensureOrderedDocsAreDeleted(sorter: Sorter): void {
+  for (const prop of Object.keys(sorter.sorts)) {
```
micheleriva, line 169: Move `Object.keys` outside the loop condition to avoid accessing the object keys on every iteration

packages/orama/src/components/sorter.ts
allevo: Yep, it is not serialize with JSON.stringify 

```
% node
Welcome to Node.js v20.2.0.
Type ".help" for more information.
> JSON.stringify({ a: new Map() })
'{"a":{}}'
> a = new Map()
Map(0) {}
> a.set('a', 0)
Map(1) { 'a' => 0 }
> JSON.stringify({ a })
'{"a":{}}'
> a
Map(1) { 'a' => 0 }
> 
```

But probably we can use a custom toJSON function to avoid this issue.

packages/orama/src/components/sorter.ts
allevo: yes correct

```
 % node
Welcome to Node.js v20.2.0.
Type ".help" for more information.
> JSON.stringify({ a: new Map() })
'{"a":{}}'
> a = new Map()
Map(0) {}
> a.set('a', 0)
Map(1) { 'a' => 0 }
> JSON.stringify({ a })
'{"a":{}}'
> a
Map(1) { 'a' => 0 }
> 
```

A custom toJSON function is required here.

micheleriva: It's quite strange that our tests couldn't identify this. We need to fix that sooner or later

H4ad: I already cover the serialization here: https://github.com/oramasearch/orama/pull/434/files/698be5fbff8958e80fa30b29465c45deca238cf9#diff-7bf96b26edf177621478ee69fd6da1265f94d1ad031a8468084ec5853ed7e35dR265-R277

packages/orama/src/components/sorter.ts
H4ad: Yep, we don't need to serialize this property.

packages/orama/src/components/sorter.ts
ShogunPanda: Correct.

ShogunPanda:
Other than things that @micheleriva outlined, looks good to me.

micheleriva:
Looks good to me. @allevo / @ShogunPanda would you be so kind to test this locally with serialization/deserialization before merging?

ShogunPanda:
@micheleriva I'm currently full on my list. @allevo can probably test this quicker.

packages/orama/src/components/sorter.ts
```diff
@@ -103,56 +101,101 @@ async function insert(
   if (!sorter.enabled) {
     return
   }
+
+  sorter.language = language
```
allevo, line 105: I don't think we want to support a language change during the insertion: the language remains the same as the DB creation.

allevo (commented):
HI! Thanks for your contribution.
I'm checking the code.
Does this change also affect the documentation?

allevo:
Another tip: could you ensure after the `insertMany` / `removeMany`the sorting?

packages/orama/src/components/sorter.ts
H4ad: This is what Orama does already: https://github.com/oramasearch/orama/blob/main/packages/orama/src/components/sorter.ts#L111

And we don't have access to Orama instance, so this could land a breaking change, so I don't think that is worth the issue of doing that.

H4ad:
@allevo We can do that, but doesn't make sense on the optimization side: https://github.com/oramasearch/orama/pull/434#issuecomment-1619698118

And this change doesn't change the documentation.

micheleriva (approved):
LGTM. Amazing job @H4ad

PR #435, opened 2023-07-04
perf: remove unused properties from radix

H4ad:
There is currently no reason for `radix` to have an `id` or `parent` property, so can we remove it?

This breaks a test in `dataset.test.ts`, but breaks the order of the `snapshot`, from what I can see the documents are just being returned in a different order than expected.

### DIFF
--- a/packages/orama/src/trees/radix.ts
+++ b/packages/orama/src/trees/radix.ts
@@ -1,12 +1,9 @@
-import { syncBoundedLevenshtein } from '../components/levenshtein.js'
-import { Nullable } from '../types.js'
-import { getOwnProperty, syncUniqueId } from '../utils.js'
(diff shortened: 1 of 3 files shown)


micheleriva:
@H4ad tests are not passing, but I can't understand how this refactor could affect pagination results

H4ad:
@micheleriva Probably is due the default sort is not stable: https://medium.com/@fsufitch/is-javascript-array-sort-stable-46b90822543f

micheleriva:
@allevo any thoughts here?

allevo:
I'm not worried about it. The sorting feature has a dedicated test suite, so it is preserved.

@H4ad  could you fix the test suite?

micheleriva (approved):
LGTM

PR #436, opened 2023-07-04
perf: replace node object for node class

H4ad:
Calling `Object.defineProperty` is not expensive but it's not cheap too ([reference](https://github.com/RafaelGSS/nodejs-bench-operations/blob/main/v20/RESULTS-v20_0_0.md#adding-property)).

So, to emulate the same behavior and have no penalty for the performance, we can just change it to a class.


H4ad:
I didn't test the serialization, but you are right, I will add a specific case to handle the `load` method for these nodes.

H4ad:
@micheleriva Actually, this is not a bug? Because when we load it back, and then serialize it again, how do you set `serialize` for those nodes that were serialized before?

H4ad:
I fixed the problem with the serialization, I didn't like the recursion, but at least it's not throwing an error with 1M items.

packages/orama/src/components/index.ts
```diff
@@ -491,9 +491,22 @@ export async function getSearchablePropertiesWithTypes(index: Index): Promise<Re
   return index.searchablePropertiesWithTypes
 }
 
+function loadNode(node: RadixNode): RadixNode {
+  const convertedNode = radixCreate(node.end, node.subWord, node.key, );
+
+  convertedNode.docs = node.docs;
+  convertedNode.word = node.word;
+
+  for (const childrenKey of Object.keys(node.children)) {
```
micheleriva, line 500: Why not `const childrenKey in node.children`? AFAIK that should be a bit faster

packages/orama/src/components/index.ts
```diff
@@ -502,6 +515,20 @@ export async function load<R = unknown>(raw: R): Promise<Index> {
     fieldLengths,
   } = raw as Index
 
+  const indexes: Index['indexes'] = {};
+
+  for (const prop of Object.keys(rawIndexes)) {
```
micheleriva, line 520: Same here, maybe a `for..in` loop here would help performances

micheleriva (approved):
Two minor things, but LGTM. @allevo would appreciate your feedback before merging

allevo (approved):
The performance remains almost the same, but the code will be clearer.
Could you address Michele's suggestions?

LGTM.

packages/orama/src/components/index.ts
H4ad: I don't know exactly if is slower or faster.

For [this benchmark](https://measurethat.net/Benchmarks/Show/2375/0/for-in-vs-objectkeys#latest_results_block), `for in` is faster.
For [this other](https://simone-sanfratello.github.io/public/node-bench-iteration/index.html), it is slower.

The only thing I know is `for in` usually causes `deoptmizations` on v8, refs: 
- https://github.com/zhangchiqing/OptimizationKillers/blob/master/README.md
- https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#5-for-in

micheleriva: In your benchmark:

```js
for (var i=10000; i > 0; i--) {
  Object.keys(obj).forEach(key => console.log(key));
}
```

there's also an extra function call as a callback for the `forEach` method. This adds slight overhead, so I don't think the benchmark itself is reliable enough.

> https://github.com/zhangchiqing/OptimizationKillers/blob/master/README.md
> https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#5-for-in

These links are interesting and I'd rather follow them, lgtm

PR #437, opened 2023-07-05
fix: issue with max call stack when search

### LINKED ISSUE #301: 200k documents search leads to maximum call stack exceeded
**Describe the bug**
When searching with Lyra for 200k documents in it it throws the error `Maximum call stack exceeded`. This is error is thrown exactly starting from 100,188 documents matching the search query.

**To Reproduce**

Execute the code below.

```
import { create, insertBatch, search } from '@lyrasearch/lyra';

async function main() {
  const db = await create({
    schema: {
      title: 'string',
    },
  });

  // the search works exactly until a document count of 100187
  const data = Array.from({ length: 200_000 }).map(() => ({ title: 'tttt' }));

  await insertBatch(db, data);

  const res = await search(db, {
    term: 'tttt',
    properties: ['title'],
  });

  console.log('hits count', res.hits.length);
}

main();

```

**Expected behavior**
Search result should be returned by lyra without throwing an error.

 - OS: macOS Ventura 13.2.1
 - ts-node version 10.9.1 and tauri version 1.2.3

H4ad:
Closes #301 

I just have a simple issue with the `.push`.

I added a function instead of just the if statement because probably `orama` can have other parts with the same issue.

PS: To be able to test/run the script on #301 without taking several minutes/hours, merge this branch with #434.

### DIFF
--- a/packages/orama/src/components/groups.ts
+++ b/packages/orama/src/components/groups.ts
@@ -1,6 +1,6 @@
 import type { Orama, ScalarSearchableValue, TokenScore, GroupByParams, GroupResult, Result, Reduce } from '../types.js'
 import { createError } from '../errors.js'
-import { getNested, intersect } from '../utils.js'
+import { getNested, intersect, safeArrayPush } from '../utils.js'
 import { getDocumentIdFromInternalId } from './internal-document-id-store.js'
 
 interface PropertyGroup {
@@ -169,7 +169,11 @@ function calculateCombination(arrs: ScalarSearchableValue[][], index = 0): Scala
   const combinations = []
   for (const value of head) {
     for (const combination of c) {
-      combinations.push([value, ...combination])
+      const result = [value];
+
+      safeArrayPush(result, combination)
+
+      combinations.push(result)
     }
   }
 
--- a/packages/orama/src/components/index.ts
+++ b/packages/orama/src/components/index.ts
@@ -34,7 +34,7 @@ import {
   Node as RadixNode,
   removeDocumentByWord as radixRemoveDocument,
 } from '../trees/radix.js'
-import { intersect } from '../utils.js'
+import { intersect, safeArrayPush } from '../utils.js';
 import { BM25 } from './algorithms.js'
 import { getInnerType, getVectorSize, isArrayType, isVectorType } from './defaults.js'
 import {
@@ -462,7 +462,7 @@ export async function searchByWhereClause<I extends OpaqueIndex, D extends Opaqu
       }
 
       const filteredIDs = idx[operation.toString() as keyof BooleanIndex]
-      filtersMap[param].push(...filteredIDs)
+      safeArrayPush(filtersMap[param], filteredIDs);
       continue
     }
 
@@ -477,8 +477,7 @@ export async function searchByWhereClause<I extends OpaqueIndex, D extends Opaqu
         const term = await context.tokenizer.tokenize(raw, context.language, param)
         for (const t of term) {
           const filteredIDsResults = radixFind(idx, { term: t, exact: true })
-          filtersMap[param].push(...Object.values(filteredIDsResults).flat())
-        }
+          safeArrayPush(filtersMap[param], Object.values(filteredIDsResults).flat())}
       }
 
       continue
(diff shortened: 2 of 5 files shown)


allevo:
Hi!
Thanks for this contribution! Will this affect the search performance when the result count is less than the limit?
How the value `10_000` is chosen?

H4ad:
@allevo I describe more the reason here: https://github.com/oramasearch/orama/pull/437/files#diff-eeb84b4af585397ddd81a97513980b6a568f6b3e82a0a597effba24233873a40R13-R18

But essentially, was just the max value divided by 10.

About the search time, from what I see, the performance for more than 10K could be increased but not too much, both operations are very performant compared to the time that will take to search that amount of items.

Also, push is more performance than concat: https://www.measurethat.net/Benchmarks/Show/4223/0/array-concat-vs-spread-operator-vs-push

allevo:
I make some benchmarks for this PR, which apparently introduces a performance gain.

I used this code:
```ts
import { create as createDev, insert as insertDev } from "orama-dev"
import { create as createStable, insert as insertStable } from "@orama/orama"
import { Benchmark } from "kelonio"
import { create as createDev, insert as insertDev } from "orama-dev"
import { create as createStable, insert as insertStable } from "@orama/orama"
import { faker } from '@faker-js/faker'

async function createSearch() {
  const n = 30000
  const data1 = Array.from({ length: n }).map(() => faker.string.sample())
  const data2 = [...data1]

  const benchmark = new Benchmark()
  const db1 = await createStable({ schema: { name: "string" } })
  for (const name of data1) {
    await insertStable(db1, { name })
  }  
  
  const db2 = await createDev({ schema: { name: "string" } })
  for (const name of data2) {
    await insertDev(db2, { name })
  }  

  await benchmark.record("search - stable", async () => await insertStable(db1, { name: data1.pop() }), {
    iterations: n,
  })
  await benchmark.record("search - dev", async () => await insertDev(db2, { name: data2.pop() }), {
    iterations: n,
  })
  printResult(benchmark)
}


function printResult(benchmark: Benchmark) {
  const data = benchmark.data
  const names = Object.keys(data)
  const results = names.map(name => {
    const result = data[name]
    const mean = result.durations.reduce((a, b) => a + b, 0) / result.durations.length

    return { name, mean, totalDuration: result.totalDuration }
  })

  results.sort((a, b) => a.mean - b.mean)

  console.log(
    results.map(result => `${result.name}: ${result.mean} ms`).join("\n")
  )
  console.log('\n\n\n')
}
```

Prints
```
search - stable: 1.754566665599999 ms
search - dev: 2.1923525156000085 ms
```

allevo:
Sorry my bad. I posted the wrong test:
```ts

async function createSearch() {
  const n = 30000
  const data1 = Array.from({ length: n }).map(() => faker.string.sample())
  const data2 = [...data1]

  const benchmark = new Benchmark()
  const db1 = await createStable({ schema: { name: "string" } })
  for (const name of data1) {
    await insertStable(db1, { name })
  }  
  
  const db2 = await createDev({ schema: { name: "string" } })
  for (const name of data2) {
    await insertDev(db2, { name })
  }  

  await benchmark.record("search - stable", async () => await insertStable(db1, { name: data1.pop() }), {
    iterations: n,
  })
  await benchmark.record("search - dev", async () => await insertDev(db2, { name: data2.pop() }), {
    iterations: n,
  })
  printResult(benchmark)
}
```

H4ad:
Hey @allevo, do you have any suggestions on this change?

allevo:
Something like
```ts
export function safeAddNewItems<T>(arr: T[], newArr: T[]) {

  if (newArr.length < MAX_ARGUMENT_FOR_STACK) {
    arr.push(...newArr)
  } else {
    for (let i = 0; i < newArr.length; i += MAX_ARGUMENT_FOR_STACK) {
      arr.push(...newArr.slice(i, i + MAX_ARGUMENT_FOR_STACK))
    }
  }
}
```

I didn't tested the performance differences.

rawpixel-vincent:
Hi, 
I've got this issue and tested this PR, it fixes my issues but I had to use `safeAddNewItems` in two more place inside of `searchByWhereClause` for my search query to work.

Attached the patch I use against 1.0.10 (also have a fix in removeDocument that crash when the node doesnt exists, part of updateMultiple)
[@orama+orama+1.0.10.patch](https://github.com/oramasearch/orama/files/12092348/%40orama%2Borama%2B1.0.10.patch)

packages/orama/src/components/groups.ts
```diff
@@ -168,7 +168,11 @@ function calculateCombination(arrs: ScalarSearchableValue[][], index = 0): Scala
   const combinations = []
   for (const value of head) {
     for (const combination of c) {
-      combinations.push([value, ...combination])
+      const result = [value];
+
+      safeArrayPush(result, combination)
+
+      combinations.push(result)
```
micheleriva, line 175: I fear I missed something; do we need both `safeArrayPush(result, combination)` and `combinations.push(result)` here? Otherwise, LGTM

allevo:
@H4ad  Could you align this branch to the main? I'm curious to see the difference in the performance.

packages/orama/src/components/groups.ts
H4ad: Yes, because `result` is just `[value, ...combination]`.

So we need to push it.

allevo:
This PR introduces an overhead:
```
search string - stable: 2.092553788 ms
search string - dev: 2.301789926000001 ms
```

I wonder if we want to fix the issue: your search is very short term. In fact, the problem is raised only when an index returns more than 100K elements. In that case, an error is thrown.

H4ad:
Orama gets the IDs of all documents to `filter` and `sort`, then, it limits that list by the number of items the user wants, usually `10`.

Now, for every list that push more than 10K elements, we batch push instead of push all those items to avoid issue with max call stack.

micheleriva:
I think that looks good, @H4ad could you solve conflicts before merging? Thanks!

H4ad:
@micheleriva Done!

micheleriva (approved):
LGTM

micheleriva:
> This PR introduces an overhead:
> 
> ```
> search string - stable: 2.092553788 ms
> search string - dev: 2.301789926000001 ms
> ```
> 
> I wonder if we want to fix the issue: your search is very short term. In fact, the problem is raised only when an index returns more than 100K elements. In that case, an error is thrown.
> 
> @micheleriva WDYT?

Hi @H4ad, I'm a bit concerned about this performance degradation. I'm not sure it's worth it to add such a overhead for this edge case

H4ad:
@micheleriva The overhead is about 0.3ms for 30k items, for most of the cases, it will not even be noticed.

I think worth the change since is better to be a little bit slower than crash.

H4ad:
@allevo I think we can increase, @micheleriva can you do more tests with different values of `MAX_ARGUMENT_FOR_STACK`, it just need to be less than 100K.

H4ad:
Accords [to this link](https://phuoc.ng/collection/this-vs-that/concat-vs-push/), webkit has a lower limit, 65k items.

allevo:
It's OK to put the magic number to 65K: this allows not to throw and limits the performance issue to large results searches.

micheleriva:
Let's use 65k as limit then. I'll merge this as soon as the conflicts are fixed

micheleriva (approved):
LGTM

PR #438, opened 2023-07-07
feat: first draft of id shortener

marco-ippolito:
first possible implementation of id-shortener:
the data structure is 
`
export type IdStore = {
  lastShort: string
  originalToShort: Record<string, string>
  shortToOriginal: Record<string, string>
}`

it will start generating ids from `A, B, C` saving the reference generated to original and other way around, so every id is stored 2 times . I chose this solution because is very easy to serialize, but I'm sure there might be something way better.
If this idea is valid I'll work on the integration with the tree

### DIFF
--- a/packages/orama/src/id-shortener.ts
+++ b/packages/orama/src/id-shortener.ts
@@ -1,6 +1,6 @@
 import { IdStore } from './types.js'
 
-const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*()'
+const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789!#$%&()*+,-./:;<=>?@[]^_`{|}~';
 
 export function generateShortId(last: string) {
   if (last === '') {
@@ -21,8 +21,10 @@ export function generateShortId(last: string) {
 }
 
 export function generateUniqueShortId(db: IdStore, original: string): string {
-  if (Object.hasOwn(db.originalToShort, original)) {
-    return db.originalToShort[original]
+
+  const existing = db.originalToShort[original]
+  if (existing) {
+    return existing
   }
 
   const newId = generateShortId(db.lastShort)
--- a/packages/orama/src/id-shortener.ts
+++ b/packages/orama/src/id-shortener.ts
@@ -1,0 +1,54 @@
+import { IdStore } from './types.js'
+
+const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*()'
+
+export function generateShortId(last: string) {
+  if (last === '') {
+    return characters[0]
+  }
+
+  const lastGeneratedIdLastIndex = last.length - 1
+
+  const lastChar = last[lastGeneratedIdLastIndex]
+
+  const charIndex = characters.indexOf(lastChar)
+
+  if (charIndex === characters.length - 1) {
+    return last + characters[0]
+  }
+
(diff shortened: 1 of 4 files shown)


packages/orama/src/id-shortener.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  if (charIndex === characters.length - 1) {
+    return last + characters[0]
+  }
+
+  return last.substring(0, lastGeneratedIdLastIndex) + characters[charIndex + 1]
+}
+
+export function generateUniqueShortId(db: IdStore, original: string): string {
+  if (Object.hasOwn(db.originalToShort, original)) {
```
allevo, line 24: Why not `db.originalToShort[original]` ?

packages/orama/src/id-shortener.ts
```diff
@@ -0,0 +1,54 @@
...
+  return db.shortToOriginal[id]
+}
+
+export function getGeneratedId(db: IdStore, id: string) {
+  return db.originalToShort[id]
+}
+
+export function createNewIdDatabase() {
+  return {
+    lastShort: '',
```
allevo, line 49: This could be an array of integers: the integers are the indexes of `characters`. In this way, the implementation of `generateShortId` should be more straightforward.

Something like:
```ts
function buildString(indexes: number[]) {
  return indexes.map(i => characters[i]).join('')
}

function updatePreviousIndex(indexes: number[], i: number) {
  if (number === -1) // Not possible {
    return
  }

  if (indexes[i] === characters.length) {
    updatePreviousIndex(indexes, i - 1)
  }

  number[i] ++
}

export function generateShortId(indexes: number[]) {
  if (indexes[indexes.length-1] === characters.length) {
    updatePreviousIndex(indexes, indexes.length - 1)
    indexes.push(0)
  }

  return buildString(indexes)
}
```

I don't know if it is more readable. WDYT?

packages/orama/src/id-shortener.ts
```diff
@@ -0,0 +1,54 @@
+import { IdStore } from './types.js'
+
+const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*()'
```
allevo, line 3: We can use safely also:
- `=`
- `£` // Not sure
- `[`
- `]`
- `{`
- `}`
- `-`
- `+`
- `,`
- `.`
- `;`
- `:`
- `?`
- `|`
- `_`
- `~`

This reduces the string length.

packages/orama/src/id-shortener.ts
marco-ippolito: Great idea!! This saves a lot of space 🙌

marco-ippolito: Actually I'm not sure it would make any difference

allevo: You are right. Too complex logic without any reason. OOOPS

allevo:
Hi Marco. Recently we merged this PR https://github.com/oramasearch/orama/pull/441, in somehow overlaps with this PR.

marco-ippolito:
> Hi Marco. Recently we merged this PR https://github.com/oramasearch/orama/pull/441, in somehow overlaps with this PR.

Have a look @H4ad, I think this allows even smaller ids

H4ad:
@marco-ippolito Sorry for the delay, our implementations are very similar but I decided to choose numbers instead of strings, so I don't think we can get smaller IDs than a number.

But your implementation can be used on https://github.com/oramasearch/orama/blob/f5aa6c238536c9f108f34ff122312fdc6d0f795a/packages/orama/src/utils.ts#L113-L115

Probably it will generate a smaller ID than the current implementation.

But this will require you to store that information somewhere, I suggest you move that logic to `internal-document-id-store.ts

micheleriva:
Closing this due to inactivity

PR #441, opened 2023-07-09
feat: id as number

### LINKED ISSUE #426: Introduce Orama internal ID for documents
# Introduction to the problem

As per the Orama documentation ([link](https://docs.oramasearch.com/usage/insert#custom-document-ids)):

> Orama automatically uses the id field of the document, if found.
>
> That means that given the following document and schema:
> ```js
> import { create, search } from '@orama/orama'
> 
> const db = await create({
>   schema: {
>     id: 'string',
>     author: 'string',
>     quote: 'string',
>   },
> })
>
> await insert(db, {
>   id: '73cbcc79-2203-49b8-bb52-60d8e9a66c5f',
>   author: 'Fernando Pessoa',
>   quote: "I wasn't meant for reality, but life came and found me",
> })
>```
>
> The document will be indexed with the following `id`: `73cbcc79-2203-49b8-bb52-60d8e9a66c5f`.
>
> If the `id` field is not found, Orama will generate a random `id` for the document.

This gives users a great opportunity to use their own custom IDs, but at a cost: Orama uses several data structures (AVL Trees, Radix Trees, Inverted Indexes) where the ID of a document gets duplicated multiple times.

For example, if we insert the following document:

```js
{
  id: '37fd6bb7-5ac2-4a37-adcd-485045b54bdc',
  author: 'Michele',
  quote: 'Hello, world!'
}
```

The id `37fd6bb7-5ac2-4a37-adcd-485045b54bdc` will get duplicated at least three times:

1. `'Michele'` will be stored in the radix tree, where the last node will contain the reference of the document containing this specific token
2. `'Hello'` will be stored in the radix tree, where the last node will contain the reference of the document containing this specific token
3. `'world'` will be stored in the radix tree, where the last node will contain the reference of the document containing this specific token

There might be other places where this ID can get duplicated depending on certain conditions.

# How to solve

If the user is indexing the content using a `UUID`, this will drastically affect the index size, because of several duplications of a quite large string. 

Therefore, we should let users index every document with their own custom IDs (without putting them into a Radix Tree, therefore making the `id` property not searchable), but use an internal, shorter ID, possibly using the `syncUniqueId` function exported [here](https://github.com/oramasearch/orama/blob/main/packages/orama/src/utils.ts#L92) to store the document reference in our data structures.

Users should be able to retrieve their docs using the `getById` function ([link](https://docs.oramasearch.com/usage/utilities#getbyid)), but internally, Orama should always use a short, optimized ID.

# Bounty Program

This issue is subject to our **Open Source Bounty Program**, and we'll reward whoever is creating a PR that gets merged with $800 for this activity.

H4ad:
/claim #426

This is my attempt to closes #426, in this implementation, I added a new component called `internalDocumentIDStore` which is responsible for keeping the internal IDs and also has a list to reverse those IDs.

- [x] Added a new ID Store
- [x] Added support for all methods to handle the internal ID
- [x] Keep the API Support with plugins (instead of sending the internal ID, I only send the original ID)
- [x] Add support for serialization and deserialization (I didn't make it backward compatible if you want this, let me know)
- [x] Fixed all tests

 About the performance, to compare this change with the old behavior, here are the current stats for the old behavior:

> To be able to run this quickly, I rebase using #434 otherwise this will take a lot of time.

```
insert: 884.5955260000192ms
memory: 184.7719955444336MB
save: 172.09495500009507ms
total size: 42.089162826538086 MBs
```

And the new behavior is:

```
insert: 798.1346069998108ms
memory: 187.97183227539062MB
save: 132.9290320002474ms
total size: 27.622851371765137 MBs
```

> To check this new behavior locally, checkout to this branch: https://github.com/h4ad-forks/orama/tree/feat/id-as-number-faster

<details>
<summary>database-size.mjs</summary>

```js
import { create, insert, save } from './dist/index.js';
import { writeFileSync } from 'fs';
import crypto from 'crypto';

(async () => {
  const db = await create({
    schema: {
      name: "string"
    }
  });

  let now = performance.now();
  for (let i = 0; i < 1e5; i++) {
    await insert(db, {
      id: crypto.randomUUID(),
    });
  }
  console.log(`insert: ${performance.now() - now}ms`);
  console.log(`memory: ${process.memoryUsage().heapUsed / 1024 / 1024}MB`);
  now = performance.now();

  const rawState = await save(db);
  console.log(`save: ${performance.now() - now}ms`);

  const jsonState = JSON.stringify(rawState);
  const totalSize = jsonState.length;
  console.log(`total size: ${totalSize / 1024 / 1024} MBs`);

  writeFileSync('./database-size.json', jsonState, 'utf8');
})();
```
</details

We had a little increase in memory usage (1.7%) but we decreased the index size by 34.36%!

This is what looks like the current serialization:

```json
{
   "internalIdStore":{
      "internalIdToId":[
         "4dca7125-6c6f-461c-9cb0-b0dba12119bc"
      ]
   },
   "index":{
      "indexes":{
         "name":{
            "word":"",
            "subWord":"",
            "children":{
               "b":{
                  "word":"ba2c536",
                  "subWord":"ba2c536",
                  "children":{
                     
                  },
                  "docs":[
                     1
                  ],
                  "end":true
               }
            },
            "docs":[
               
            ],
            "end":false
         }
      },
      "searchableProperties":[
         "name"
      ],
      "searchablePropertiesWithTypes":{
         "name":"string"
      },
      "frequencies":{
         "name":{
            "1":{
               "ba2c536":1
            }
         }
      },
      "tokenOccurrencies":{
         "name":{
            "ba2c536":1
         }
      },
      "avgFieldLength":{
         "name":1
      },
      "fieldLengths":{
         "name":{
            "1":1
         }
      }
   },
   "docs":{
      "docs":{
         "1":{
            "id":"4dca7125-6c6f-461c-9cb0-b0dba12119bc",
            "name":"ba2c536"
         }
      },
      "count":1
   },
   "sorting":{
      "sortableProperties":[
         "name"
      ],
      "sortablePropertiesWithTypes":{
         "name":"string"
      },
      "sorts":{
         "name":{
            "docs":{
               "1":0
            },
            "orderedDocs":[
               [
                  1,
                  "ba2c536"
               ]
            ],
            "type":"string"
         }
      },
      "enabled":true,
      "isSorted":true
   }
}
```


allevo:
Hi @H4ad ! Could you resolve the conflicts?

Thanks

packages/orama/src/components/algorithms.ts
```diff
@@ -16,16 +17,24 @@ export function prioritizeTokenScores(arrays: TokenScore[][], boost: number, thr
     for (let j = 0; j < entriesLength; j++) {
       const [token, score] = arr[j]
       const boostScore = score * boost
+      const oldScore = tokenMap.get(token)
```
allevo, line 20: If I remember correctly, get has very bad performance when the key is not found. It is better to use `has` and then `get`

packages/orama/src/components/algorithms.ts
```diff
@@ -1,12 +1,13 @@
...
 import { TokenScore, BM25Params } from '../types.js'
+import { InternalDocumentID } from './internal-document-id-store.js';
 
 export function prioritizeTokenScores(arrays: TokenScore[][], boost: number, threshold = 1): TokenScore[] {
   if (boost === 0) {
     throw createError('INVALID_BOOST_VALUE')
   }
 
-  const tokenMap: Record<string, number> = {}
+  const tokenMap = new Map<InternalDocumentID, number>()
```
allevo, line 10: are we able to build this map with some content?
For instance, `pop()` from the right entries list and use that to build the map initially.
This could reduce the `get` failure

packages/orama/src/components/documents-store.ts
```diff
@@ -1,54 +1,64 @@
...
+
+  if (typeof store.docs[internalId] === 'undefined') {
     return false
   }
 
-  delete store.docs[id]
+  delete store.docs[internalId]
   store.count--
 
   return true
```
allevo, line 64: we should remove also from `sharedInternalDocumentStore`, right?

packages/orama/src/methods/search.ts
```diff
@@ -204,6 +204,7 @@ export async function search<AggValue = Result[]>(
 
   // Get unique doc IDs from uniqueDocsIDs map
   let uniqueDocsArray = Object.entries(context.uniqueDocsIDs)
+    .map(([id, score]) => [+id, score] as TokenScore)
```
allevo, line 207: why this is required?

packages/orama/src/methods/search.ts
```diff
@@ -223,6 +224,7 @@ export async function search<AggValue = Result[]>(
       uniqueDocsArray = docsWithIdAndScore.map(([id, score]) => [id, score])
     } else {
       uniqueDocsArray = await orama.sorter.sortBy(orama.data.sorting, uniqueDocsArray, params.sortBy)
+        .then(results => results.map(([id, score]) => [getInternalDocumentId(orama.internalDocumentIDStore, id), score]))
```
allevo, line 227: I don't like this map here. Why not put it inside sortBy?

packages/orama/src/methods/search.ts
```diff
@@ -246,7 +248,12 @@ export async function search<AggValue = Result[]>(
   }
 
   if (typeof results !== 'undefined') {
-    searchResult.hits = results.filter(Boolean)
+    for (const result of results) {
+      if (!result) continue
+
+      result.id = getDocumentIdFromInternalId(orama.internalDocumentIDStore, +result.id)
+      searchResult.hits.push(result)
```
allevo, line 255: This `push` increases the size of the result allocating again the vector. Is it really needed?

packages/orama/src/components/internal-document-id-store.ts
```diff
@@ -0,0 +1,70 @@
+import { Orama } from '../types.js';
+
+export type DocumentID = string | number
```
allevo, line 3: Why allow both?

packages/orama/src/methods/search.ts
```diff
@@ -317,7 +324,7 @@ async function fetchDocumentsWithDistinct(
       continue
     }
 
-    results.push({ id, score, document: doc! })
+    results.push({ id: id.toString(), score, document: doc! })
```
allevo, line 327: Is this really needed?

packages/orama/src/methods/search.ts
```diff
@@ -360,7 +367,7 @@ async function fetchDocuments(
       // We retrieve the full document only AFTER making sure that we really want it.
       // We never retrieve the full document preventively.
       const fullDoc = await orama.documentsStore.get(docs, id)
-      results[i] = { id, score, document: fullDoc! }
+      results[i] = { id: id.toString(), score, document: fullDoc! }
```
allevo, line 370: also here

packages/orama/tests/group.test.ts
```diff
@@ -1,4 +1,5 @@
 import t from 'tap'
+import { inspect } from "util";
```
allevo, line 2: remove this

allevo (commented):
This PR seems ok. I left some comments. In particular, related to DocumentID: why do we need to support both, number or string?

packages/orama/src/components/algorithms.ts
H4ad: IDs could be duplicated, so I don't think populating this map first will be faster than just leaving it empty.

image

This is the behavior when keys are duplicated.

packages/orama/src/components/documents-store.ts
H4ad: If we just remove it from `sharedInternalDocumentStore.idToInternalId`, it's fast enough.

If we try to remove it from `sharedInternalDocumentStore.internalIdToId`, it will be slower as `sorter`.

What we can do is perform a cleanup of `internalIdToId` on serialize.

What do you prefer?

packages/orama/src/methods/search.ts
H4ad: It will require me to change the `.sortBy` to return just `InternalID`, I didn't want to change because of https://github.com/oramasearch/orama/pull/441#discussion_r1261368165

But if you are okay with introducing this breaking change, I can change and it will not require anymore the `.map`.

packages/orama/src/methods/search.ts
H4ad: Not really, I will just modify the `.id` and then set the `.hits = results`.

packages/orama/src/methods/search.ts
allevo: keep like that.

packages/orama/src/components/internal-document-id-store.ts
allevo: I'm ok to introduce a little breaking change with this. Regarding the naming. Is this a documented or IndexId?

micheleriva: This will bump Orama to `v1.1.0`

packages/orama/src/components/documents-store.ts
allevo: No sorry, we should implement it.
For now, we don't care about the performance during the remove.

micheleriva: > For now, we don't care about the performance during the remove.

Agreed. I'd dedicate a separate PR to this

packages/orama/src/components/internal-document-id-store.ts
H4ad: Ok, so I will change all the references for `DocumentID` to `InternalID`, and I will only use `DocumentID` on `getByID` and when we return the documents from `search`.

> Regarding the naming. Is this a documented or IndexId?

I didn't understand your question but `DocumentID` is just an alias to reference the ID of the Document that was generated or was passed by the user, we you see this type, is referring to these two cases.

But I think we should add some documentation about it on Orama Docs, just to be clear about how we store IDs to be more efficient.

allevo: Ok, understood you point

allevo (approved):
Again, amazing work!

LGTM

micheleriva (approved):
Terrific job @H4ad. As always!

LGTM

Issue #445, opened 2023-07-15
Support defining a db with jsonschema


mrbrianevans:
**Is your feature request related to a problem? Please describe.**
JSON schema is a widely used and understood format for defining the shape of data structures. Existing applications already make use of it and have schemas written for entities. It would make adoption of orama search easier if a database could be created using a json schema rather than only the custom schema DDL that orama currently offers.

**Describe the solution you'd like**
```ts
import { create } from '@orama/orama'
 
const movieDB = await create({
  jsonschema: {
    type: 'object',
    properties: {
      title: { type: 'string' },
      director: { type: 'string' },
      plot: { type: 'string' },
      year: { type: 'number' },
      isFavorite: { type: 'boolean' }
    },
    required: [ 'title', 'director', 'plot', 'year', 'isFavorite' ]
  },
})
```

**Describe alternatives you've considered**
Writing my own custom script to try and convert my existing json schemas into oramas schema format.

**Additional context**
JSON schema website: https://json-schema.org/
There is widely available tooling available for json schema, and generating json schema from almost any language. 
Consider the case of a very complex ecommerce site with a Golang backend, all the entities are defined in Go structs and stored in Mongodb. They want frontend in-memory search. They would need to rewrite all their schemas in oramas format to make the fields searchable. 
With my feature request, they could simply run their Go structs through a jsonschema generator such as https://github.com/invopop/jsonschema and then all fields are instantly searchable with no manual schema writing.

**Additional benefit**
Not the main feature request but a side benefit is that typescript types can be inferred from jsonschema, meaning the database could be automatically typed after being created with a jsonschema, so `insert`, `update` and `search` could all have the correct types without being specified by the user, since it gets inferred from the jsonschema.

nxf-gerdjungbluth:
We just had a team discussion about this feature request and really liked it. Maybe we will be able to contribute.

mrbrianevans:
Based on @allevo 's feedback, I suggest something like this:
```ts
import { create } from '@orama/orama'
import { schemaFromJson } from '@orama/jsonschema'
 
const movieDB = await create({
  schema: schemaFromJson({
    type: 'object',
    properties: {
      title: { type: 'string' },
      director: { type: 'string' },
      plot: { type: 'string' },
      year: { type: 'number' },
      isFavorite: { type: 'boolean' }
    },
    required: [ 'title', 'director', 'plot', 'year', 'isFavorite' ]
  }),
})
```
[...]

PR #452, opened 2023-07-21
[plugin]: orama schema from json schema

ilteoood:
Possible solution for #445.

This is an initial basic solution, just to understand if orama wants this kind of plugin.

Todo for the future: support `$ref`


packages/schema/src/jsonSchema.ts
```diff
@@ -0,0 +1,45 @@
...
+    return ORAMA_SUPPORTED_TYPES.has(jsonSchema.type) || isArraySupportedByOrama(jsonSchema);
+}
+
+const extractOramaType = (jsonSchema: JSONSchema4): SearchableType => {
+    const oramaType = ORAMA_SUPPORTED_TYPES.has(jsonSchema.type) ? jsonSchema.type : `${(jsonSchema.items as JSONSchema4)!.type}[]`
+
+    return oramaType as SearchableType
+}
+
+export const schemaFromJson = (jsonSchema: JSONSchema4) => {
```
allevo, line 31: For now, the $ref is not supported, and it is ok. Anyway, if someone in the future wants to implement it, we should allow async operation. For instance, we need to allow the download of external references.
I prefer to move this to async function.

allevo (commented):
Amazing!
Thanks, this is very helpful!

Only one note left

packages/schema/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@orama/schema",
```
allevo, line 2: @orama/plugin-json-schema

packages/schema/src/jsonSchema.ts
```diff
@@ -0,0 +1,45 @@
...
+    assertTypeObject(jsonSchema)
+
+    const oramaSchema: Schema = {}
+
+    for (const [propertyName, propertyDefinition] of Object.entries(jsonSchema.properties || {})) {
+        if (isSupportedByOrama(propertyDefinition)) {
+            oramaSchema[propertyName] = extractOramaType(propertyDefinition)
+        } else if (isJsonObject(propertyDefinition)) {
+            oramaSchema[propertyName] = schemaFromJson(propertyDefinition)
+        }
```
allevo, line 41: should we throw an error here?

ilteoood:
@allevo I'll be on holiday until next week, I'll address all of them once I'll be back!

packages/schema/package.json
ilteoood: I was thinking to keep it generic so if in the future you would like to generate the schema from other sources, like yaml or toml, you don't have lots of packages to maintain

packages/schema/src/jsonSchema.ts
ilteoood: Does orama throws if it receive a wrong schema? If not, I would not throw here either

ilteoood: Looking at the `validateSchema` function provided by Orama, it's ignoring the invalid types, I would do the same here

PR #453, opened 2023-07-22
plugin: orama schema from json schema

mrbrianevans:
Converts from json schema to orama schema format.

Sorry I didn't see that someone else had already opened a PR until after I'd implemented it.

Closes https://github.com/oramasearch/orama/issues/445.

Very simple implementation, doesn't resolve references. That can be done by many other packages on npm, this is just to convert from a fully resolved json schema.

### DIFF
--- a/packages/plugin-jsonschema/README.md
+++ b/packages/plugin-jsonschema/README.md
@@ -1,0 +1,13 @@
+# Nextra Plugin
+
+[![Tests](https://github.com/oramasearch/orama/actions/workflows/turbo.yml/badge.svg)](https://github.com/oramasearch/orama/actions/workflows/turbo.yml)
+
+Official plugin to convert schemas in the jsonschema format to work with Orama.
+
+# Usage
+
+For the complete usage guide, please refer to the [official plugin documentation](https://docs.oramasearch.com/plugins/plugin-nextra).
+
+# License
+
+[Apache-2.0](/LICENSE.md)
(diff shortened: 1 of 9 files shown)


packages/plugin-jsonschema/test/index.test.ts
```diff
@@ -0,0 +1,131 @@
...
+  t.plan(4)
+
+  t.test('throw error for nested arrays', t => {
+    t.plan(1)
+    const jsonschema = {
+      type: 'object',
+      properties: {
+        coOrdinates: {
+          type: 'array',
+          items: { type: 'array', items: { type: 'number' } },
```
allevo, line 95: This si supported as 
```ts
{
  coOrdinates: 'number[]'
}
```

Or I missing something?

packages/plugin-jsonschema/README.md
```diff
@@ -0,0 +1,13 @@
+# Nextra Plugin
```
allevo, line 1: wrong title

ilteoood:
@allevo this is a duplicate of #452, just to know: which one would you like to merge?

packages/plugin-jsonschema/test/index.test.ts
mrbrianevans: Its a doubly nested array of arrays, data would look like this:
```
"coOrdinates": [
  [5, 7],
  [8, 3]
]
```
As far as I know orama doesn't support indexing that, so the converter to JSONschema would throw an error if it was attempted.

PR #417, opened 2023-06-09
Failing threshold.test.ts

kilbot:
I believe this test will fail with r4.count = 3.

PR #454, opened 2023-07-24
feat: Improved threshold handling.

ShogunPanda:
This PR reworks threshold handling to make sure it works as intended.

Fixes and supersedes #417.

### DIFF
--- a/packages/orama/src/components/algorithms.ts
+++ b/packages/orama/src/components/algorithms.ts
@@ -2,12 +2,18 @@ import { createError } from '../errors.js'
 import { TokenScore, BM25Params } from '../types.js'
 import { InternalDocumentID } from './internal-document-id-store.js'
 
-export function prioritizeTokenScores(arrays: TokenScore[][], boost: number, threshold = 1): TokenScore[] {
+export function prioritizeTokenScores(
+  arrays: TokenScore[][],
+  boost: number,
+  threshold = 1,
+  keywordsCount: number,
+): TokenScore[] {
   if (boost === 0) {
     throw createError('INVALID_BOOST_VALUE')
   }
 
-  const tokenMap = new Map<InternalDocumentID, number>()
+  const tokenScoresMap = new Map<InternalDocumentID, number>()
+  const tokenKeywordsCountMap = new Map<InternalDocumentID, number>()
 
   const mapsLength = arrays.length
   for (let i = 0; i < mapsLength; i++) {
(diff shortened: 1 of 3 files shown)


packages/orama/src/components/algorithms.ts
```diff
@@ -41,16 +49,47 @@ export function prioritizeTokenScores(arrays: TokenScore[][], boost: number, thr
...
-    return results.slice(0, shortestArrayLength)
+    return results.slice(0, lastTokenWithAllKeywords + 1)
   }
 
   // If the threshold is between 0 and 1, we will return all the results that contains at least the threshold of search terms
   // For example, if threshold is 0.5, we will return all the results that contains at least 50% of the search terms
   // (fuzzy match with a minimum threshold)
-  const thresholdLength = Math.ceil((threshold * 100 * results.length) / 100)
+  const thresholdLength =
+    lastTokenWithAllKeywords + Math.ceil((threshold * 100 * (results.length - lastTokenWithAllKeywords)) / 100)
```
allevo, line 92: Are there some tests that test this? In particular, the comment above.

ShogunPanda: Yup, see: https://github.com/oramasearch/orama/blob/improved-sort/packages/orama/tests/threshold.test.ts#L128

packages/orama/src/components/algorithms.ts
```diff
@@ -41,16 +49,47 @@ export function prioritizeTokenScores(arrays: TokenScore[][], boost: number, thr
...
+  const allResults = results.length
+  const tokenKeywordsCount: [InternalDocumentID, number][] = []
+
+  for (const tokenKeywordsCountEntry of tokenKeywordsCountMap.entries()) {
+    tokenKeywordsCount.push(tokenKeywordsCountEntry)
+  }
+
+  // Find the index of the last result with all keywords.
+  // Note that since score is multipled by 1.5 any time the token is encountered in results it means
+  // that tokenMap and tokenKeywords should always have the same order.
```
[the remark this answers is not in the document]
ShogunPanda: Yeah, I renamed the variable and forgot about the comment. Fixing it.

About the addition below: later we actually sort by score and number of keywords, so this doesn't really matter.
My only point here is to emphasize that if a document has a significantly bigger score than another, it means it appeared in more keywords search than the other and thus `tokenScoresMap` and `tokenKeywordsCountMap` have the same key order if we order by values.

allevo (approved):
LGTM

micheleriva (approved):
LGTM

PR #458, opened 2023-08-02
Implement enum type

allevo:
We collect some issues for our performance during the insertion. In particular, the number type has some problems if the number of documents is high. But also string can be slow too. 

The issue related to the insertion time is because of the cost of keeping the order for the number. Trees allow us to keep order with the cost of O(log(N)) for insertion time. But what happens if we choose to lose the order to have a faster insertion time?

Depending on what you need, values can have a lot of meanings:
- strings you want to search for
- numbers you want to compare with operators (`>`, `<`, `>=`, `<=`, `=`, `!=`)
- enum you want to compare with operators (`=`, `in`)

This PR introduces the `enum` type to allow:
- fast insertion
- fast search
- `=` and `in` operator (others??)
- reduce the dump size

What this PR misses:
- [x] tests: done
- [x] we need to identify `FlatTree` properly: used a discriminator string
- [x] choose if we want to support `enum[]` as well: discarded for now
- [x] documentation: done
- [x] collect suggestions and comment: done

Usage:
```ts
const db = await create({
  schema: {
    categoryId: 'enum',
  },
})

const [c2, c3, c5] = await insertMultiple(db, [
  { categoryId: 1 },
  { categoryId: 1 },
  { categoryId: 2 },
  { categoryId: 3 },
  { categoryId: "5" },
])
const resultEq = await search(db, {
  term: '',
  where: {
    categoryId: { eq: 1 },
  }
}) // Expected 2 results
const resultIn = await search(db, {
  term: '',
  where: {
    categoryId: { in: [1, 3, "5"] },
  }
}) // Expected 4 results

```

### DIFF
--- a/packages/docs/pages/usage/create.mdx
+++ b/packages/docs/pages/usage/create.mdx
@@ -18,7 +18,7 @@ Orama supports the following types:
 | `string`         | A string of characters.                                                     | `'Hello world'`                                                             |
 | `number`         | A numeric value, either float or integer.                                   | `42`                                                                        |
 | `boolean`        | A boolean value.                                                            | `true`                                                                      |
-| `enum`           | An enum value.   (since version 1.3.0)                                      | `'drama'`                                                                   |
+| `enum`           | An enum value.                                                              | `'drama'`                                                                   |
 | `string[]`       | An array of strings.                                                        | `['red', 'green', 'blue']`                                                  |
 | `number[]`       | An array of numbers.                                                        | `[42, 91, 28.5]`                                                            |
 | `boolean[]`      | An array of booleans.                                                       | `[true, false, false]`                                                      |
--- a/packages/docs/pages/usage/create.mdx
+++ b/packages/docs/pages/usage/create.mdx
@@ -4,12 +4,7 @@ import { Callout } from 'nextra-theme-docs'
 
 We can create a new instance (from now on database) with an **indexing `schema`**.<br/>
 The schema represents **the searchable properties** of the document to be inserted.
-
 Not all properties need to be indexed, but only those that we want to be able to search for.
-
-Adding a property to the schema will make it searchable, but it will also increase the size of the database and overall insertion and search times.
-
-Therefore, you should only index the properties that you need to search for.
 
 If you want to learn more and see real-world examples, check out [this blog post](https://oramasearch.com/blog/optimizing-orama-schema-optimization) we wrote about schema optimization.
 
@@ -23,10 +18,11 @@ Orama supports the following types:
 | `string`         | A string of characters.                                                     | `'Hello world'`                                                             |
 | `number`         | A numeric value, either float or integer.                                   | `42`                                                                        |
 | `boolean`        | A boolean value.                                                            | `true`                                                                      |
+| `enum`           | An enum value.   (since version 1.3.0)                                      | `'drama'`                                                                   |
 | `string[]`       | An array of strings.                                                        | `['red', 'green', 'blue']`                                                  |
 | `number[]`       | An array of numbers.                                                        | `[42, 91, 28.5]`                                                            |
 | `boolean[]`      | An array of booleans.                                                       | `[true, false, false]`                                                      |
-| `vector[<size>]` | A vector of numbers to perform vector search on.                            | `[0.403, 0.192, 0.830]`                                                               |
+| `vector[<size>]` | A vector of numbers to perform vector search on.                            | `[0.403, 0.192, 0.830]`                                                     |
 
 A database can be as simple as:
 
@@ -74,6 +70,18 @@ const movieDB = await create({
     isFavorite: 'boolean',
   },
 })
+
+await insert(db, {
+  title: 'The Godfather',
+  plot: 'The aging patriarch of an organized crime dynasty transfers control of his clandestine empire to his reluctant son.',
+  cast: {
+    director: 'Francis Ford Coppola',
+    leading: 'Marlon Brando',
+    supporting: ['Al Pacino', 'James Caan', 'Robert Duvall'],
+  },
+  year: 1972,
+  isFavorite: true,
+})
 ```
 
 ## Vector properties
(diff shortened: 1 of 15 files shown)


packages/orama/src/trees/flat.ts
```diff
@@ -0,0 +1,81 @@
...
+import { EnumComparisonOperator, ScalarSearchableType, ScalarSearchableValue } from "../types.js"
+
+export interface FlatTree {
+  foo: string
+  numberToDocumentId: Map<ScalarSearchableValue, InternalDocumentID[]>
+}
+
+export function create(): FlatTree {
+  return {
+    foo: 'bar',
```
ShogunPanda, line 11: Why do you need this?

ShogunPanda:
I'm not sure I understand the purpose of this. Can you please clarify?

allevo:
Hi,
sorry for the delay.
I'll try to describe the proposal in a better way.
Sorry for the wall.

# Context
Now, Orama uses tree structures to store strings and numbers. In particular, it uses radix trees and avl trees to store respectively strings and numbers.

# Problem
The time insertion is high when the developer trying to insert more than 100k of documents.
Some users opened issues or report that issues on slack.
In particular, the insertion inside avl trees took times.

We uses trees because that structures have some guarantees like:
- find all the documents mapped by a defined value
- find all documents greater or lesser that a value
- for radix tree, find all documents using a tolerance

All those features are great for some kind of data:
- strings where you want to find a substring
- filter for number equity
- filter for number using a comparator

All those searches (and others) are yet supported by Orama.

Anyway, if I think generally a search, I'm thinking a search bar + filters. Meanwhile the search bar completely leverages on radix tree for typo tolerance for instance, the filters are typically implemented using two kind of interfaces:
- the final user can specify a range of a value. The example is the price or the distance from the city center.
- the final user can specify the presence or the absence of a specific value inside the document. For instance, if an hotel has a services like a parking, or if a book has the Kindle version, or t-shirt size.

For the first use case, we can leverage on avl because we need the comparators.
For the second use case, we *could* leverage on the same implementation but it creates an overhead because we are still keeping the order also for fields where the order or the typo tolerance don't make any sense.

# Solution
I'm proposing a different approach to not keep the order for fields that doesn't require it. Like the example above, the user still want to filter on it with an exact match and without keeping an order.
This solution offers a way to implement it using a `Map<any, InternalDocumentId[]>`.
Benefits:
- 30% improvement for number insertion
- less memory usage (not yet tested)
- reduce the dump size (not yet tested)

Const:
- add another type that cannot so easy to understand.


The implementation is still in progress because I would like to understand if this makes sense before finalizing it.

allevo:
Hi!
What do you mean by a new component? Which component are you expecting to have?
It is always an index, right?

packages/orama/src/trees/flat.ts
allevo: At the searching time, we don't have the `schema`. So the logic to differentiate the index is made on the "filter" type.
I mean, a tree is cast to the boolean tree if the value of the filter is a boolean. The same is true for the string. In the end, as a fallback, there's the number logic.
That should be different: we should keep the kind of tree inside the tree itself and use that value to cast to the right type.
This is a simple trick to differentiate the AVL tree from this tree, so differentiate number filters from enum ones.
Is it correct? Surely not.

Any suggestions?

ShogunPanda: Can't you access the schema through the Orama object?

allevo:
What do you mean? I want to add this feature to the default index.
If this is not the case, could you give me an example of your thoughts?

ShogunPanda:
I think that in order to keep Orama clean things should be separated. Nothing prevents you to implement the new feature as new index and then call within the default index.
But I still think this should be kept as separated component.

packages/orama/src/components/index.ts
```diff
@@ -67,9 +75,23 @@ export type VectorIndex = {
   }
 }
 
+export type Tree = {
```
micheleriva, line 78: Please move these unions into separate type definitions, for example (pseudocode):

```ts
type TreeType =
  | 'avl'
  | 'radix'
  | 'bool'
  | 'flat'

type TTree<T = TreeType, N = unknown> = {
  type: T,
  node: N
}

export type Tree = 
  | TTree<'radix', RadixNode>
  | TTree<'avl',   AVLNode<number, InternalDocumentID[]>>
  | TTree<'bool',  BooleanIndex>
  | TTree<'flat',  FlatTree>
```

packages/orama/src/trees/flat.ts
```diff
@@ -0,0 +1,79 @@
+import { InternalDocumentID } from "../components/internal-document-id-store.js"
+import { EnumComparisonOperator, ScalarSearchableValue } from "../types.js"
+
+export interface FlatTree {
+  numberToDocumentId: Map<ScalarSearchableValue, InternalDocumentID[]>
+}
+
+export function create(): FlatTree {
+  return {
+    numberToDocumentId: new Map()
```
micheleriva, line 10: Does this support JSON serialization?

packages/orama/src/trees/flat.ts
```diff
@@ -0,0 +1,79 @@
...
+export function insert(root: FlatTree, key: ScalarSearchableValue, value: InternalDocumentID): FlatTree {
+  if (root.numberToDocumentId.has(key)) {
+    root.numberToDocumentId.get(key)!.push(value)
+    return root
+  }
+  root.numberToDocumentId.set(key, [value])
+  return root
+}
+
+export function find(root: FlatTree, key: ScalarSearchableValue): InternalDocumentID[] | null {
```
micheleriva, line 23: Use `Nullable<>` when returning nullable values

packages/orama/src/trees/flat.ts
```diff
@@ -0,0 +1,79 @@
...
+  return root.numberToDocumentId.get(key) ?? null
+}
+
+export function remove(root: FlatTree | null, key: ScalarSearchableValue): FlatTree | null {
```
micheleriva, line 27: Use `Nullable<>` when declaring nullable arguments

packages/orama/src/trees/flat.ts
```diff
@@ -0,0 +1,79 @@
...
+    throw new Error('Invalid operation')
+  }
+
+  const operationType = operationKeys[0] as keyof EnumComparisonOperator
+  switch (operationType) {
+    case 'eq': {
+      const value = operation[operationType]!
+      return root.numberToDocumentId.get(value) ?? []
+    }
+    case 'in': {
```
micheleriva, line 65: Maybe we want to support `nin` (not in) as well?

micheleriva (commented):
Looks good to me, just a couple of minor changes. Will approve once we also have documentation ready. Well done!

packages/docs/pages/usage/create.mdx
```diff
@@ -23,10 +18,11 @@ Orama supports the following types:
 | `string`         | A string of characters.                                                     | `'Hello world'`                                                             |
 | `number`         | A numeric value, either float or integer.                                   | `42`                                                                        |
 | `boolean`        | A boolean value.                                                            | `true`                                                                      |
+| `enum`           | An enum value.   (since version 1.3.0)                                      | `'drama'`                                                                   |
```
micheleriva, line 21: I would remove "(since version 1.3.0)". We also introduced vectors from v1.2.0 but the docs should always reflect the latest release (or else, it will soon become a mess)

packages/orama/src/components/index.ts
```diff
@@ -67,9 +75,28 @@ export type VectorIndex = {
   }
 }
 
+export const enum TreeType {
```
micheleriva, line 78: Don't use enums in TypeScript. You don't have control over their compilation in JavaScript.

micheleriva (commented):
I think it mostly looks good, except for a couple of small comments. @allevo please add tests for Enums for the `plugin-data-persistence` package as well.

Good to go as soon as these comments are addressed.

Terrific job 🔥

micheleriva (approved):
LGTM

PR #462, opened 2023-08-03
Adds vector search capabilities to Orama

micheleriva:
This PR introduces vector search capabilities in Orama. This will be part of Orama `v1.2.0`, the next major release.

## New Vector API

Orama adds support for Vector search by adding a new datatype: `vector[<size>]`:

```js

const db = await create({
  schema: {
    text: 'string',
    embedding: 'vector[1536]' // <--- vector size is mandatory here. OpenAI embeddings, for instance, have 1536 dimensions. 
  }
})
```

After you create your Orama instance, you can insert and search for vectors using the new `searchVector` function:

```js
import { create, insert, searchVector } from '@orama/orama'

const db = await create({
  schema: {
    text: 'string',
    myVector: 'vector[5]'
  }
})

await insert(db, { text: 'foo', myVector: [1, 0, 0, 0, 0] })
await insert(db, { text: 'bar', myVector: [0, 0, 0, 0, 0] })
await insert(db, { text: 'baz', myVector: [1, 1, 1, 1, 1] })

const results = await searchVector(db, {
  vector: [1, 0, 0, 0, 0], // Your input vector
  property: 'myVector' // Property to search through is mandatory with the "searchVector" function
})
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ the
 
 # Highlighted features
 
-- [Vector Search](https://docs.oramasearch.com/usage/search/vectors)
+- [Vector Search](https://docs.oramasearch.com/usage/search/vector-search)
 - [Search filters](https://docs.oramasearch.com/usage/search/filters)
 - [Facets](https://docs.oramasearch.com/usage/search/facets)
 - [Fields Boosting](https://docs.oramasearch.com/usage/search/fields-boosting)
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 </h4>
 <br />
 <p align="center">
-  A resilient, innovative and open-source search experience to achieve <br />
+  A resilient, innovative and open-source full-text and vector search experience to achieve <br />
   seamless integration with your infrastructure and data
 </p>
 <br />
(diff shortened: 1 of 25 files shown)


gustavopch:
Awesome addition! One question: have you tested the performance? Like, how much time would it take to search among 30k vectors, 100k vectors, etc.? 🤔

micheleriva:
Hey @gustavopch, thank you so much!

I share your concerns; I made some tests locally and found no significant performance problems. I am sure though that we will find some edge cases and will eventually implement something similar to the gist you shared if we see that it's becoming more and more challenging to scale performances.

With that being said, I'd like to highlight a couple of things:

1. You may not need OpenAI embeddings (since you use 1536 dimensions, I guess you're talking about them). Let me share this great article by Supabase: https://supabase.com/blog/fewer-dimensions-are-better-pgvector
2. Cosine similarity is incredibly easy to parallelize. We plan to write some platform-specific plugins (Node.js, browsers, etc.) to parallelize execution on large vector sets.

Hope that helps :)

PR #470, opened 2023-08-27
feat: added strong typed orama schema

### LINKED ISSUE #456: I am unable to specify a data type for my index, and that compromises the type safety
**Describe the bug**
I am using Orama search in a NestJS project. The use case is as follows:

- When a module initializes (e.g. UserModule), in the onModuleInit() method a call to the database is made to retrieve all available users.
- The data is indexed in an Orama index, to be made available for searching later and avoid unnecessary calls to the DB.
- In case a new user is created or an existing one is updated, respective calls are made to the index to keep the data up to date.

The issue I'm facing is that I can't find a way to specify a type for the objects this index will be storing. Since the schema is only used to specify the searchable fields, I feel there should be a way to tell the index the full data schema that it will be holding. Currently, the search function returns the hits and the documents that were hit, but the return type is Document and it isn't generic. This causes some issues for other services (e.g. UserService), which are required to return an array of users - User[]. 

The current workaround is to either:
- cast the document to unknown, and then to the type you need
`const user = document as unknown as User`

- manually map each property you need:
```typescript
  private toUser(document: Document): User {
    return {
      id: doc['id'] as string,
      name: doc['name'] as string,
      email: doc['email'] as string,
      ...
    };
  }
```

**To Reproduce**
1. Create a simple service that holds an Orama index:

```typescript
@Injectable()
export class UserSearchService {
  private index: Orama;

  constructor(private dbService: DbService) {}

  async initializeIndex(): Promise<void> {
    this.index = await create({
      schema: {
        id: 'string',
        name: 'string',
        email: 'string',
      },
    });

    const users = await this.dbService.findAll<User>();
    await insertMultiple(this.index, users, 500);
  }

  async find(): Promise<User[]> {
    const result = await search(this.index, {...});

    const users = result.hits.map((hit) => hit.document);
    return users;
  }
```

2. You will get a type error:

```
Type 'Document[]' is not assignable to type 'User[]'.
  Type 'Record<string, unknown>' is missing the following properties from type 'User': id, name, email
```

3. Casting will also not work properly and will give the following error:

```
Conversion of type 'Document' to type 'User' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
```

4. The same errors occur the other way, when you are trying to update the index or insert something in it - for example, your service will receive an User object and you will not be able to insert it because it isn't of type Document.

**Expected behavior**
For the sake of type safety, I would expect to be able to specify the form of data my index will be holding, and avoid casting and / or mapping. A good approach would be to make all Orama methods generic, or at least the create method - so you can pass the type of your index when making it.

Example given:
```typescript
const index = await create<User>({ // this will specify the full form of the data
     // and this will specify the searchable fields
      schema: {
        id: 'string',
        name: 'string',
        email: 'string',
        isActive: 'boolean',
        issuer: 'string',
        roles: 'string[]',
      },
    });

// this could either be generic or not, but the type specified at creation would guarantee the type safety of the returned document
const user = await search<User>(index);    
```

**Additional context**
The version of Orama I am facing this issue on is 1.1.1

H4ad:
Closes #456
/claim #456

This PR turns the Orama completely type-safe, below some examples of how to use:

## Schema

Now, if you want to have the type from the result of your schema, you can do the following:

```ts
import { Schema } from '@orama/orama'

const mySchema = {
  title: 'string'
} as const;
type MySchemaDocument = Schema<typeof mySchema>;
```

Be careful with `as const`, it is required to be able to correctly infer your schema, if you didn't make it const, your document will be inferred to `never`.

## Create

By default, create will create the correct typings from schema automatically:

```ts
const db = await create({
  schema: {
    quote: 'string',
    author: 'string',
  },
})
```

The return type will be:

image

It also works for more complex schemas:

image

If you didn't customize the index, sorter, or documentStore, it will infer to the default implementations.

## Insert & Update

During the insert, all the properties are optional and you can also specific custom properties, like:

```ts
const oldDocId1 = await insert(db, {
  quote: "Life is what happens when you're busy making other plans",
  author: 'John Lennon',
  customProp: 'Olá'
})
```

But, if you specify a wrong value for some value in the schema, for example passing a number to quote, you will receive an error:

image

## Search

With search, you will have two behaviors, by default the result type will be the typed-schema:

image

But, in cases where the result will be too much different from the typed-schema, you can do the following:

image

Also, now we automatically suggest options for `groupBy.properties`, `sortBy.property` and `where`, like this:

image

image

Currently, nested properties such as `doc.title` are not suggested even if you have schema support, but you can specify them without problems.

## Sharing Typed Orama Instance

In cases where you need the type for Orama, you can do the following:

```ts
import { Schema, TypedDocument } from '@orama/orama'

const mySchema = {
  title: 'string'
} as const;

type MyOrama = Orama<typeof mySchema>;

// the following will be the same as Schema<typeof mySchema>;
type MySchemaDocument = TypedDocument<MyOrama>
```

## Hooks

The hooks from `components` also are correctly typed based on the schema:

image

But is not possible to infer extra properties, like when your schema has just `title` but in your document you also insert `description`.

In those cases, you should do the following:

```ts
const db = await create({  
  schema: {
    strings: 'string[]',
    num: 'number[]',
    b: 'boolean[]',
  },
  components: {
    beforeInsert: (orama, key, doc) => {
      type MyDoc = { strings: string[], num: number[], b: boolean[], extraProperty: string };
      const typedDoc = doc as unknown as MyDoc
    }
  }
})
```

But nothing stops you from just grab the property you want, like:

```ts
console.log(doc!.extraProperty);
```

It just will just be typed as `any`.

## Breaking Changes

I needed to upgrade the typescript to 5.0.0 because of the new modifier `const`, this makes it easier for the users to specify the schema without the need of doing `{ schema: { title: 'string' } as const }`.


micheleriva:
Hi @H4ad , thanks! Would you be able to fix conflicts?

micheleriva (commented):
I think this mostly looks good to me. I think we will need to update the docs as well if we want to use `as const` when defining the schema. But I really with we could get rid of it actually... so we can avoid such a drastic change.

@H4ad will this work on TypeScript <5?

@allevo could you also take a look at this?

H4ad:
@micheleriva I'm still working on types, and cleaning the tests (some of them passes but the types is actually broken).

It could work on <5 but I don't think we will get the same results and API that we have now since `create` will not be able to infer correctly the types of the schema.

H4ad:
Now the code is stable enough, I also added better types for search (I missed some opportunities to correctly type), so now I think the types are good enough.

allevo:
If I understood correctly, this PR propagates the schema definition to the insert/insertMultiple/search methods: amazing!
We should mention something in the documentation.

allevo (changes requested):
We should update the documentation as well
Something like "orama is fully typed so the document type is inferred from the schema" in create, insert, and search pages.

WDYT?

H4ad:
@allevo I have no skills to write good documentation about this, I think the documentation should be in another PR as we did for internal IDs, since this is a breaking change and also very important, this probably will deserve a blog post too.

allevo (approved):
ok I'll care about docs! No problem.

LGTM

packages/docs/pages/usage/typescript.mdx
```diff
@@ -0,0 +1,49 @@
...
+```typescript copy
+await insert(db, {
+  title: 'Big Fish',
+  director: 'Tim Burton',
+  plot: 'Will Bloom returns home to care for his dying father, who had a penchant for telling unbelievable stories. After he passes away, Will tries to find out if his tales were really true.',
+  year: 2004,
+  isFavorite: true
+})
+const s = await search(db, { term: 'fish' })
+const year: any = s.hits[0].document.year // not inferred, so it's any
```
H4ad, line 48: It's not inferred but you can make it type-safe by passing the interface with that field, like: `search<WithYearField>`

allevo: good point. I'm changing it!

H4ad: The tests that you have on Orama don't test for types, you can literally break the types and the tests will not break, they only break when the assert is false (at least that was the behavior I had when I was changing the Orama tests).

But I can add it by the end of the day.

allevo: Yeah, till now, we don't have that kind of test. But with this PR, type types are guaranteed. So, a little test on that would be amazing!
At least, we document this capability in the test suite also.

H4ad: In this case, we can use: https://github.com/SamVerschueren/tsd

Is the same lib used by https://github.com/sindresorhus/type-fest/blob/main/test-d/abstract-class.ts

H4ad: But probably this will require a new project and command to test only the types, separated from the default tap tests.

allevo: `tsd` seems to do the job greatly.
I'll go for just a new command run after the test suite. it should work.

packages/docs/pages/index.mdx
```diff
@@ -19,40 +19,6 @@ Orama is a fast, batteries-included, full-text and vector search engine entirely
 
 Get started with just a few lines of code:
```
micheleriva, line 20: @allevo please remove this line as well

micheleriva (approved):
LGTM. Just a little fix for @allevo

micheleriva:
Hi @H4ad, we only need a few conflict fixes to get this one merged. Thanks!

H4ad:
@micheleriva @allevo what should be the type for enum? string, number? both?

H4ad:
Based on the tests, I leave it as `string | number`.

And we can merge this PR before #482, I don't want to resolve conflicts for the third time :sweat:

micheleriva:
@H4ad you're right. Sorry if it took so long for us to merge, it's a pretty complex one. Thanks for your work!

allevo:
Thanks!

PR #471, opened 2023-08-30
Add `saveWithHighlight` and `loadWithHighlight`

### LINKED ISSUE #446: searchWithHighlight crashes after persistence restore
**Describe the bug**
When db is restored using data persistence plugin the search with higlight functionality no longer works

**To Reproduce**

- Add presistence plugin
- Add match highlight plugin
- Insert documents
- call persist
- call restore
- call searchWithHighlight

**Expected behavior**
not crash with error

**Screenshots**
```
const db = await create({
        schema: {
            // id: 'string',
            text: 'string',
            name: 'string',
            type: 'string',
            year: 'number',
            source: 'string',
            author: 'string'
        },
        components: {
            // Register the hook
            afterInsert: [highlightAfterInsert],
            tokenizer: {
                stopWords: englishStopwords,
                stemming: true,
                stemmer
            },
        }
    })
```
```
TypeError: Cannot read properties of undefined (reading '21650872-55')          1:49:20 PM
    at file:///Users/trinimac/WebstormProjects/lazarites/node_modules/@orama/plugin-match-highlight/dist/index.js:52:78
    at Array.map (<anonymous>)
    at searchWithHighlight (file:///Users/trinimac/WebstormProjects/lazarites/node_modules/@orama/plugin-match-highlight/dist/index.js:51:30)
    at async cachedEventHandler.swr (file:///Users/trinimac/WebstormProjects/lazarites/.nuxt/dev/index.mjs:5363:21)
    at async file:///Users/trinimac/WebstormProjects/lazarites/.nuxt/dev/index.mjs:1584:20
    at async _resolve (file:///Users/trinimac/WebstormProjects/lazarites/.nuxt/dev/index.mjs:1445:23)

```

 - NodeJS Nuxt 3 server

benjaminleonard:
See here for context: https://github.com/oramasearch/orama/issues/446#issuecomment-1693672265

New to the project so apologies if this isn't the right approach. It's extending the native `save()` and `load()` since `"positions"` is not present and included in either. It does the raise question of how this would work if you had multiple plugins that needed to extend the base `Orama` object?

Another approach could be by passing an array of the attributes you wanted the base `save()` and `load()` to include.

Fixes #446

### DIFF
--- a/packages/plugin-match-highlight/src/commonjs.cts
+++ b/packages/plugin-match-highlight/src/commonjs.cts
@@ -4,7 +4,9 @@ import type { Orama } from '@orama/orama'
 import type {
   afterInsert as esmAfterInsert,
   OramaWithHighlight,
-  searchWithHighlight as esmSearchWithHighlight
+  searchWithHighlight as esmSearchWithHighlight,
+  saveWithHighlight as esmSaveWithHighlight,
+  loadWithHighlight as esmLoadWithHighlight
   // @ts-expect-error Ignore broken resolution - This errors when using tsconfig.cjs.json
 } from './index.js'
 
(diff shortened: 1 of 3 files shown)


micheleriva (approved):
That's awesome, thanks! LGTM

PR #474, opened 2023-08-31
Tokenize skip properties

philippebeaulieu:
I want to used facet search but many of the values we have are multi-word-token (ie. "black matte") that needs to be searched without tokenization, not only stemming but no split at all.

I added a new optionnal parameter that replicate the stemmingSkipProperties that was already in place.

If a property is matched, then no splitting occurs on the matching prop.


micheleriva (approved):
LGTM. I'm gonna merge this, but would you also be able to open a separate PR for updating the docs?

Thanks!

PR #481, opened 2023-09-15
fix(plugin-nextra): add missing cmd+k keybinding to focus the input search

stearm:
As the title says, I noticed that the `cmd+k` shortcut is missing.


micheleriva (approved):
Thanks! LGTM!

PR #482, opened 2023-09-18
Implements enum[]

allevo:
With this PR, I want to introduce the `enum[]` as the follower of #458

So,
```javascript
const db = await create({
  schema: {
   colors: 'enum[]',
  },
})
await insertMultiple(db, [
  {colors: ['green', 'red'] },
  {colors: ['green'] },
  {colors: ['blue', 'white'] },
])

const resultContains = await search(db, {
  term: '',
  where: {
    colors: { contains: 'green' },
  }
}) // Expected 2 results: 1° and 2°
const resultIntersects = await search(db, {
  term: '',
  where: {
    colors: { intersects: ['red', 'blue'] },
  }
}) // Expected 2 results: 1° and 3°
```

This PR introduces `contains`, and `intersects` operators because they are more appropriate for arrays.

- [x] implementation
- [x] documentation
- [x] tests
- [x] facets for `enum` and `enum[]`
- [ ] feedback

### DIFF
--- a/packages/docs/pages/usage/create.mdx
+++ b/packages/docs/pages/usage/create.mdx
@@ -22,6 +22,7 @@ Orama supports the following types:
 | `string[]`       | An array of strings.                                                        | `['red', 'green', 'blue']`                                                  |
 | `number[]`       | An array of numbers.                                                        | `[42, 91, 28.5]`                                                            |
 | `boolean[]`      | An array of booleans.                                                       | `[true, false, false]`                                                      |
+| `enum`           | An enum value.                                                              | `['drama', 'crime', 'war']`                                                 |
 | `vector[<size>]` | A vector of numbers to perform vector search on.                            | `[0.403, 0.192, 0.830]`                                                     |
 
 A database can be as simple as:
--- a/packages/docs/pages/usage/create.mdx
+++ b/packages/docs/pages/usage/create.mdx
@@ -22,7 +22,7 @@ Orama supports the following types:
 | `string[]`       | An array of strings.                                                        | `['red', 'green', 'blue']`                                                  |
 | `number[]`       | An array of numbers.                                                        | `[42, 91, 28.5]`                                                            |
 | `boolean[]`      | An array of booleans.                                                       | `[true, false, false]`                                                      |
-| `enum`           | An enum value.                                                              | `['drama', 'crime', 'war']`                                                 |
+| `enum[]`         | An array of enums.                                                          | `['comedy', 'action', 'romance']`                                           |
 | `vector[<size>]` | A vector of numbers to perform vector search on.                            | `[0.403, 0.192, 0.830]`                                                     |
 
 A database can be as simple as:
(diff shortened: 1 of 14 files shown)


packages/docs/pages/usage/create.mdx
```diff
@@ -22,6 +22,7 @@ Orama supports the following types:
 | `string[]`       | An array of strings.                                                        | `['red', 'green', 'blue']`                                                  |
 | `number[]`       | An array of numbers.                                                        | `[42, 91, 28.5]`                                                            |
 | `boolean[]`      | An array of booleans.                                                       | `[true, false, false]`                                                      |
+| `enum`           | An enum value.                                                              | `['drama', 'crime', 'war']`                                                 |
```
micheleriva, line 25: @allevo having a bad day? 😂 I would use a bit more positive examples in the docs (like `['comedy', 'action', 'romance']`)

packages/docs/pages/usage/search/filters.mdx
```diff
@@ -113,10 +113,20 @@ const results = await search(db, {
...
+
+
+## Enum[] operators
+
+The enum properties support the following operators:
+
+| Operator        | Description                        | Example                                       |
+| ---------       | ------------------------------     | ---------------------------------             |
+| `contains`      | Contains the given value           | `genre: { contains: 'drama' }`                |
+| `intersects`    | Intersects with the given array    | `genre: { intersects: ['drama', 'horror'] }`  |
```
micheleriva, line 132: Why use two different keywords? As far as I understand, `contains` is just a monomorphic version of `intersect`. We could also use `genre: { contains: ['drama', 'horror'] }` to avoid introducing an extra keyword.

packages/orama/src/components/facets.ts
```diff
@@ -76,19 +77,23 @@ export async function getFacets<S extends Schema>(
           break
         }
         case 'boolean':
+        case 'enum':
         case 'string': {
-          calculateBooleanOrStringFacet(facets[facet].values, facetValue as string | boolean, propertyType)
+          calculateBooleanStringOrEnumFacet(facets[facet].values, facetValue as string | boolean | number, propertyType)
```
micheleriva, line 82: I think it's time to move this type definition into a separate union alias

micheleriva (commented):
Looks mostly good to me. I just left a couple of small comments. Since we're operating on arrays, I would use a unique keyword `contains`, where the value is either polymorphic (string or array of strings), or monomorphic (`contains` must include an array of 1 element).

packages/docs/pages/usage/search/filters.mdx
allevo: Because contains mean a subset, not an intersection:
- the field `enum[]` is an array
- I, as a developer, provide an `array`
How can I "compare" those arrays? 
- `eq` is wrong: it depends on what means equality in this case
- `contains` means "the doc field contains the given array.". Also, this is wrong. We are not calculating the subset, just the intersection.

We can collapse it into unique `contains` but it should be implemented differently (and heavier).

micheleriva: What's the difference between:

`genre: { contains: 'drama' }`

and

`genre: { contains: ['drama'] }` ?

allevo:
Implement just `containsAll` operator

micheleriva (approved):
LGTM. In the future we might want to add another PR to implement a function for intersection

PR #491, opened 2023-09-25
add export to transformFn and defaultToSectionSchema

raiindev:
Exporting these two functions because they could be helpful for other packages based on Orama.


allevo (approved):
LGTM

Issue #212, opened 2022-12-11
 Filter results based on Geo-location field


nicksav:
**Is your feature request related to a problem? Please describe.**
Adding support to filter based on location. 

**Describe the solution you'd like**
As a minimum provide ability to filter results inside a rectangle - boundingBox

**Additional context**
I think filtering a key here. Yep, search is good, but filters, facets, geo search is highly required.

micheleriva:
Hi @nicksav,
that would be a nice feature. Let me ping as soon as we assign it a priority in our roadmap.

Thanks!

sreenath-n:
@micheleriva these two features are also very useful in most of the modern apps:
1. Filter based on center geopoint and radius (Eg. to find restaurants nearby)
2. Sort results based on distance from a geopoint (Eg. to find restaurants nearby and sort based on nearest first)

Example from `Typesense`:
```
'filter_by' : 'location:(48.90615915923891, 2.3435897727061175, 5.1 km)',
'sort_by'   : 'location(48.853, 2.344):asc'
```
`Typesense` Doc: https://typesense.org/docs/0.23.1/api/geosearch.html

W7T2A:
+1 for this, is there any update in prio?

sreenath-n:
This is the only blocker for us to switch from Typesense to Orama Search.

imlautaro:
+1 for this too

micheleriva:
Hi y'all. Still to be prioritized, will update this issue as soon as we prioritize it!

Leanvitale:
We could also use this update, since it is the only thing that prevents us from implementing Orama in our Startup.

micheleriva:
We're on it

micheleriva:
This feature is available for preview in `v2.0.0-beta.1`. Has been implemented with https://github.com/oramasearch/orama/pull/505. Please try it and give us some feedback! 🙂

PR #505, opened 2023-10-01
Implements Geosearch

micheleriva:
This PR aims to implement geolocation-based search (as requested in #212).

The initial design for the APIs is the following:

```ts
import { search, create, insert } from '@orama/orama'

const db = await create({
  name: 'string',
  position: 'geopoint'
})

await insert(db, {...})
await insert(db, {...})
await insert(db, {...})

await search(db, {
  term: 'duomo',
  where: {
    position: { // the name of the property
      radius: {
        coordinates: { lon: 11.012345, lat: 42.012345 },
        unit: 'km',
        value: 100,
        inside: false // optional, default is true
      }
    }
  }
})

await search(db, {
  term: 'duomo',
  where: {
    position: { // the name of the property
      polygon: {
        coordinates: [ { lon: 11.019283, lat: 42.019283 }, { lon: 11.0192839, lat: 42.019283 }, ...],
        inside: false // optional, default is true
      }
    }
  }
})
```


packages/orama/src/trees/bkd.ts
```diff
@@ -0,0 +1,280 @@
...
+
+export function create (points: Point[]): MaybeNode {
+  if (points.length === 0) {
+    return undefined
+  }
+
+  const stack: InsertTask[] = [{
+    points,
+    depth: 0
+  }]
```
H4ad, line 41: I suggest you initialize all the properties:

```suggestion
  const stack: InsertTask[] = [{
    points,
    depth: 0,
    parent: undefined,
    direction: undefined,
  }]
```

packages/orama/src/trees/bkd.ts
```diff
@@ -0,0 +1,280 @@
...
+
+    if ((parent != null) && direction) {
+      parent[direction] = node
+    } else {
+      root = node
+    }
+
+    if (points.slice(0, median).length > 0) {
+      stack.push({
+        points: points.slice(0, median),
```
H4ad, line 66: Since you didn't use the new array of points to return to the user, probably will be faster if you store the `start` and `end` instead of creating a new array.

packages/orama/src/trees/bkd.ts
```diff
@@ -0,0 +1,280 @@
...
+        depth: depth + 1,
+        parent: node,
+        direction: 'left'
```
H4ad, line 69: Probably using an int will be a more memory-friendly approach.

packages/orama/src/trees/bkd.ts
```diff
@@ -0,0 +1,280 @@
...
+      })
+    }
+
+    if (points.slice(median + 1).length > 0) {
```
H4ad, line 73: This is not the same as:

```suggestion
    if (median + 1 > (points.length - 1)) {
```

packages/orama/src/trees/bkd.ts
```diff
@@ -0,0 +1,259 @@
...
+    const median = Math.floor(points.length / 2)
+    const node: Node = { point: points[median] }
+
+    if ((parent != null) && direction) {
+      parent[direction] = node
+    } else {
+      root = node
+    }
+
+    if (points.slice(0, median).length > 0) {
```
H4ad, line 64: This is not the same as:

```suggestion
    if (points.length > 0) {
```

packages/orama/src/trees/bkd.ts
```diff
@@ -0,0 +1,259 @@
...
+  while (stack.length > 0) {
+    const task = stack.pop()
+    if (task == null) {
+      continue
+    }
+
+    const { points, depth, parent, direction } = task
+
+    const axis = depth % K
+    points.sort((a, b) => (axis === 0 ? a.x - b.x : a.y - b.y))
```
H4ad, line 54: Probably sorting these points first, `pointsXSorted` and `pointsYSorted`, together with the int pointers will be way faster and memory-friendly.

micheleriva:
@H4ad thanks a lot for the early review, I'm still writing/refactoring most of the code, so it's expected to see things that are not super optimized yet 🙏 

I'll make sure to ask you for a review as soon as the PR is ready :)

H4ad:
Just a little comment about a micro-optimization, probably using a tuple will use less memory than having an object to store `lat` and `lon`

micheleriva:
> Just a little comment about a micro-optimization, probably using a tuple will use less memory than having an object to store `lat` and `lon`

Yes I'm using an object during development as the property names helps me read the code easily!

allevo:
My idea for DX:
```ts
import { create, insert } from '@orama/orama'

const db = await create({
  schema: {
    name: 'string',
    city: 'string',
    center: 'geopoint[2]' // 2 dim points
  },
})
// https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.2
await insert(db, {
  name: 'Duomo',
  city: 'Milan',
  // https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.1
  center: [9.191383, 45.464211]
})

const resultsByRadius = await search(db, {
  term: 'Duomo',
  where {
    center: {
      byRadius: {
        center: [11.012345, 42.012345],
        radius: {
          unit: 'km',
          value: 100
        },
        keep: 'inner', // or 'outer'
      }
  }
})

const resultsByPolygon = await search(db, {
  term: 'Duomo',
  where {
    center: {
      byPolygon: {
        coordinates: [ [11.019283, 42.019283], [11.0192839, 42.019283 ], ...],
        keep: 'inner', // or 'outer'
      }
  }
})
```

I prefer to use:
- `geopoint[K]` to indicate which dimension we are working in
- think to use [GeoJson](https://datatracker.ietf.org/doc/html/rfc7946) over other formats.
- use an object to represent radius to avoid string parse inside Orama
- `where` where possible: we want to filter the results by other values as we did for numbers / booleans / enums etc...
- avoid `inclusive` that seems like `>` vs `>=`

micheleriva:
@allevo I reviewed your comment, I'd propose a compromise between your proposal and mine.

1. Every orama instance should have only one geo property. We can always increase this limit later.
2. You can use whatever name you want for your geo property.
3. We can use a tuple for storing longitude and latitude (as in RFC7946), even though objects give a better developer experience IMHO.
4. We don't need to specify the number of geopoints, right now we only support 2, and we default to that. Eventually, we will be able to introduce a type parameter for that.
5. Following the previous point, we might also want to support `geo[point]`, `geo[point:3]`, and `geo[multiPoint]` in the future. Maybe in a following major release.

The API might look like this:

```js
import { search, create, insert } from '@orama/orama'

const db = await create({
  name: 'string',
  coordinates: 'geopoint'
})

await insert(db, {...})
await insert(db, {...})
await insert(db, {...})

await search(db, {
  term: 'duomo',
  where: {
    geo: {
      radius: {
        coordinates: [11.012345, 42.012345],
        unit: 'km', // defaults to 'm'
        value: 100,
        inside: false // optional, default is true
      }
    }
  }
})

await search(db, {
  term: 'duomo',
  where: {
    geo: {
      polygon: {
        coordinates: [ [11.019283, 42.019283], [11.0192839, 42.019283 ], ...],
        unit: 'km', // defaults to 'm'
        value: 100,
        inside: false // optional, default is true
      }
    }
  }
})
```

That way we could reuse most of the code for polygon and radius search. What do you think?

allevo:
Another challenge.
- Return to the object annotation for coordinates: few people know it, and I would like to reduce the issue of coordinates ordering (the first is the longitude, the second is latitude, anti-intuitive IMHO). Also, typescript will support the DX, suggesting the correct type. Also, other RFCs (https://www.rfc-editor.org/rfc/rfc5870.html#section-6) use the inverse representation as a "list of numbers".
- We can implement `geopoint[2]` not for the generic `geopoint[K]` but just with that dimension:
  - we are explicitly saying we support only 2 dimension
  - we can implement the generic later.
- I don't like the `geo[point]` annotation because we can internally use different structures to index different geo types.
- I don't know if the unit should have a default. Why `m` is better than `km`? We should consider that a sort radius of the search in the geospatial world is meaningless: you can approximate it as Euclidean space. I prefer to keep the mandatory values:
  - `coordinates`
  - `unit`
  - `value`

So something like
```ts
import { search, create, insert } from '@orama/orama'

const db = await create({
  name: 'string',
  position: 'geopoint[2]'
})

await insert(db, {...})
await insert(db, {...})
await insert(db, {...})

await search(db, {
  term: 'duomo',
  where: {
    position: { // the name of the property
      radius: {
        coordinates: { lon: 11.012345, lat: 42.012345 },
        unit: 'km',
        value: 100,
        inside: false // optional, default is true
      }
    }
  }
})

await search(db, {
  term: 'duomo',
  where: {
    position: { // the name of the property
      polygon: {
        coordinates: [ { lon: 11.019283, lat: 42.019283 }, { lon: 11.0192839, lat: 42.019283 }, ...],
        unit: 'km',
        value: 100,
        inside: false // optional, default is true
      }
    }
  }
})
```

micheleriva:
@allevo looks almost good to me. I just don't like the `geopoint[2]` type, I'd really prefer just using `geopoint` right now. We will eventually support 3D worlds but will see then how to deal with them at a type level. Right now I would only focus on 2D dimensions.

I agree with coordinates ordering, but internally, we will always store them as tuples to save some space (especially when serializing the db).

I think I'll start implementing the last API design you shared, then it will be easy to find a compromise on the `geopoint` type anyway 🙂

micheleriva:
I now only need to write the geopoint deletion function, but that's overall good for review. @H4ad , @allevo

packages/orama/src/trees/bkd.ts
```diff
@@ -0,0 +1,356 @@
...
+
+  let node: Nullable<Node> = tree.root
+  let depth = 0
+
+  while (node !== null) {
+    // Check if the current node's point matches the new point
+    if (node.point.lon === point.lon && node.point.lat === point.lat) {
+      // Merge the new docIDs with the existing ones and remove duplicates
+      const newDocIDs = node.docIDs ?? []
+      node.docIDs = Array.from(new Set([...newDocIDs, ...docIDs || []]))
```
H4ad, line 56: Use `safeArrayPush`.

packages/orama/src/utils.ts
```diff
@@ -259,3 +265,22 @@ export function flattenObject (obj: object, prefix = ''): AnyDocument {
...
+    case 'm':
+      return distance
+    case 'km':
+      return distance * 1000
+    case 'ft':
+      return distance * 0.3048
+    case 'yd':
+      return distance * 0.9144
+    case 'mi':
+      return distance * 1609.344
```
H4ad, line 282: I think using object will be a little bit faster.

```suggestion
const mapDistanceToMeters = {
  cm: 0.01,
  m: 1,
  km: 1000,
  ft: 0.3048,
  yd: 0.9144,
  mi: 1609.344,
};

export function convertDistanceToMeters(distance: number, unit: GeosearchDistanceUnit): number {
  const ratio = mapDistanceToMeters[unit]
  
  if (ratio === undefined)
    throw new Error(createError('INVALID_DISTANCE_SUFFIX', distance).message)
  
  return distance * ratio
```

packages/orama/src/trees/bkd.ts
```diff
@@ -0,0 +1,356 @@
+import type { Nullable, GenericSorting } from '../types.js'
+import type { InternalDocumentID } from '../components/internal-document-id-store.js'
+
+export interface Point {
+  lon: number
+  lat: number
+}
```
H4ad, line 7: I think is more efficient and fast to store the data as an array (tuple).

packages/orama/src/types.ts
```diff
@@ -177,6 +182,34 @@ export type EnumArrComparisonOperator = {
...
+  | 'km'
+  | 'ft'
+  | 'yd'
+  | 'mi'
+
+export type GeosearchRadiusOperator = {
+  radius: {
+    coordinates: Point,
+    value: number,
+    unit?: GeosearchDistanceUnit,
```
allevo, line 197: I prefer not to use a default here.

packages/orama/src/trees/bkd.ts
micheleriva: Taking the courage to refactor now :)

H4ad (approved):

Issue #506, opened 2023-10-03
RangeError: Maximum call stack size exceeded


BanderasPRO:
**The bug**
Same as https://github.com/oramasearch/orama/pull/469 and https://github.com/oramasearch/orama/pull/437

**Steps**
1. Create a huge database from HTML
2. Then search

**Expected behavior**
No error after search

**Code**
```js
const texts = await search(ORAMA_DB, {
    term: searchTerm,
    properties: ['content'],
    where: {
      type: { nin: ['title', 'a'] },
    },
    groupBy: {
      properties: ['properties.slug'],
    },
    limit,
  });
```

**Logs**
```
RangeError: Maximum call stack size exceeded
    at filter (file:///dev/node_modules/@orama/orama/dist/trees/flat.js:78:32)
    at Object.searchByWhereClause (file:///dev/node_modules/@orama/orama/dist/components/index.js:312:43)
    at search (file:///dev/node_modules/@orama/orama/dist/methods/search.js:96:45)
```
After making a local fix, I encountered another issue:
```
RangeError: Maximum call stack size exceeded
    at Object.searchByWhereClause (file:///dev/node_modules/@orama/orama/dist/components/index.js:312:35)   
    at search (file:///dev/node_modules/@orama/orama/dist/methods/search.js:96:45)
```
Then I simply searched for `.push(...` in `@orama/orama/dist` and found a couple of places where this approach is used:
* flat.js
* sorter.js
* index.js

To be honest, I didn't clone the project; I just looked into the `node_modules`

It would be great if the fix could be released as soon as possible.

H4ad:
Hey, thanks for opening the issue. Are you willing to open a PR to fix this issue? We can ping Michele to give priority once the PR is landed.

You can use this function and change all the references for `.push` that you found:
https://github.com/oramasearch/orama/blob/13ef4c445231bda98b243dd1f87b8be7c8437c35/packages/orama/src/utils.ts#L20-L37

micheleriva:
I would really appreciate a PR for fixing this. I would definitely give this high priority

H4ad:
You should create a pull request, first fork this project, and then push your changes inside a new branch of your fork.

Then, you will see a button on the main page of your fork to create and compare a pull request against the main repo.

See more: https://docs.github.com/en/get-started/quickstart/contributing-to-projects

BanderasPRO:
PR is ready

BanderasPRO:
Hi, 
[...]

PR #507, opened 2023-10-04
fix: Using `safeArrayPush(dest, source)` instead of `dest.push(...source)` (#506)

BanderasPRO:


packages/orama/tests/insert.test.ts
```diff
@@ -4,6 +4,7 @@ import { Index } from '../src/components/index.js'
 import { getInternalDocumentId } from '../src/components/internal-document-id-store.js'
 import { AnyDocument, create, insert, insertMultiple, search } from '../src/index.js'
 import dataset from './datasets/events.json' assert { type: 'json' }
+import { safeArrayPush } from "../src/utils"
```
allevo, line 7: This is not needed. in the test, we can use `push` because the number of the objects here is not so high

allevo: Also: the build fails for this reason.

allevo:
LGTM when the CI passes.

packages/orama/tests/insert.test.ts
BanderasPRO: done

allevo (approved):
LGTM

H4ad (approved):
Awesome, thanks for the PR, and congratulations on your first PR!

micheleriva:
@H4ad thanks for your support!

PR #511, opened 2023-10-06
Strong 'boost' parameter

allevo:
Add strong type to `boost` parameter

### DIFF
--- a/packages/orama/src/types.ts
+++ b/packages/orama/src/types.ts
@@ -13,28 +13,29 @@ export type SyncOrAsyncValue<T = void> = T | PromiseLike<T>
 // Given a type T, return a new type with:
 // - the concatenation of nested properties as key
 // - the type of the nested property as value
-export type Flatten<T extends object> = object extends T
+type Flatten<T extends object> = object extends T
   ? object
   : {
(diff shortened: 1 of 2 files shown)


packages/orama/src/types.ts
```diff
@@ -236,7 +239,7 @@ export type SearchParams<T extends AnyOrama, ResultDocument = TypedDocument<T>>
   /**
    * The properties of the document to search in.
    */
-  properties?: '*' | FlattenSchemaProperty<T>[]
+  properties?: '*' | FlattenSchemaProperties<T>
```
micheleriva, line 242: Why converting this to plural form?

allevo: Because it is an array of FlattenSchemaProperty.
Do you prefer another convention?

micheleriva: Yes by convention it is better to use singular and convert them to arrays when used in lists

allevo: done!

allevo: done!

micheleriva (approved):

Issue #480, opened 2023-09-14
Bug: tolerance option not behaving as hoped


ttillberg:
Thanks for the amazing lib and clear documentation!  I'm looking at using Orama to search local chat messages (typically involving a few words up to several sentences). 

Using `@orama/orama ^1.2.3` I'm getting fast a correct results for exact and prefixed matching however however typos don't seem to work the way I was hoping. I'm probably missing the obvious but testing the `tolerance` parameter against an example in the docs returns poor results. So I'm wondering what could be wrong. 

Looking at the following example.
https://docs.oramasearch.com/usage/search/introduction#typo-tolerance

If I grab a slightly bigger database:
https://github.com/erik-sytnyk/movies-list/blob/master/db.json

```ts
{ 
  term: "Christopher Nolan", 
  properties: ["director"] 
}

// result: OK: matches 1 exact result like expected
```

```ts
{
  term: "Cris",
  properties: ["director"],
}

// result: OK: matches 1 document "Michael Cristofer" (no tolerance was set, so this is kind of expected)
```

```ts
{
  term: 'Cris',
  properties: ['director'],
  tolerance: 1,
}

// result: "fails": matches 0 documents, in the documentation this query would return all "Chris's" - not this would still fail bumping the tolerance level
// one example in the DB: "director": "Pierre Coffin, Chris Renaud",
```

here's my playground (all output is in the console):
https://codesandbox.io/p/sandbox/keen-knuth-9wql22?file=/src/main.ts:65,28

I've played with other options, such as the tokenizer, stemming, relevance, threshold but without luck. What am I missing?

micheleriva:
I fear that's a known issue. We're performing the Levenshtein edit distance on words living in the same prefix bucket, rather than performing the edit distance calculation on trees. For instance, searching for `Chris` and `hris` will give you totally different results, as they don't share a common prefix.

I'll be putting a bounty on this bug, thanks for opening it!

/bounty 500

ogil7190:
@ttillberg is this open to work?

micheleriva:
@ogil7190 yes

SP321:
@micheleriva the documentation does not seem to contain a max limit for tolerance. I am assuming a max tolerance limit of 3 per token will be considered a proper solution. can you confirm?

micheleriva:
Fixed with v1.2.11

PR #516, opened 2023-10-10
Fix string tolerance option (#480)

SP321:
/claim #480

Added a testcase for the (issue #480). 
Confirmed bug by testcase fail.
Added Levenshtein distance calculation on radix tree.
distance on tree is only calculated if exact option is not used and tolerance>0. (existing logic is used if tolerance is 0 or exact is true)
Confirmed all testcases(including added testcase of #480) passed.

### DIFF
--- a/packages/orama/src/methods/create.ts
+++ b/packages/orama/src/methods/create.ts
@@ -87,7 +87,7 @@ function validateComponents<OramaSchema, TIndex, TDocumentStore, TSorter, TOrama
 }
 
 export async function create<
-  const OramaSchema,
+  OramaSchema,
   TIndex = IIndex<Index>, 
   TDocumentStore = IDocumentsStore<DocumentsStore>, 
   TSorter = ISorter<Sorter>> ({
--- a/packages/orama/tests/search.test.ts
+++ b/packages/orama/tests/search.test.ts
@@ -6,6 +6,38 @@ t.test('search method', t => {
   t.test('with term', async t => {
     const [db, id1, id2, id3, id4] = await createSimpleDB()
 
+    //https://github.com/oramasearch/orama/issues/480
+    //following testcase pass only if issue 480 is fixed.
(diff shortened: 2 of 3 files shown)


packages/orama/tests/search.test.ts
```diff
@@ -6,6 +6,38 @@ t.test('search method', t => {
...
+            stemming: true,
+            stopWords: englishStopwords,
+          },
+        },
+      })
+
+      await insert(db, { name:"Chris "})
+      await insert(db, { name:"Craig"})
+      await insert(db, { name:"Chxy"}) //create h node in radix tree.
+      await insert(db, { name:"Crxy"}) //create r node in radix tree.
```
micheleriva, line 29: Can you try by adding:

```js
await insert(db, { name: "Dhris" })
```

and see if it matches `"Chris"` when tolerance is set to `1`?

SP321: `     await insert(db, { name: "Dhris" })
      const result1 = await search(db, { term: 'Chris', tolerance: 1 })
      const result2 = await search(db, { term: 'Cgris', tolerance: 1 })
      t.equal(result1.count, 1)
      t.equal(result2.count, 0)// should only match on tolerance 2.
`
It works as expected.

micheleriva: Can you please update the PR with this test?

SP321: > Can you please update the PR with this test?

updated PR.

micheleriva (approved):
LGTM. Thanks!

packages/orama/src/methods/create.ts
```diff
@@ -87,7 +87,7 @@ function validateComponents<OramaSchema, TIndex, TDocumentStore, TSorter, TOrama
 }
 
 export async function create<
-  const OramaSchema,
```
H4ad, line 90: @micheleriva This change will break the inference of schemas

PR #520, opened 2023-10-16
fix: create custom tokenizer

### LINKED ISSUE #492: Normalization cache overriden when creating orama instance
**Describe the bug**

When creating an orama instance, I am using the tokenizer config property to provide a normalization cache with preset values but doesn't end up being used and defaults to an empty cache. 

Unless it is intentionally ignored, the fix for this seems straightforward. ([updating this LOC](https://github.com/oramasearch/orama/blob/main/packages/orama/src/components/tokenizer/index.ts#L145)). 

**To Reproduce**

Here's a codesandbox with reproducible code. https://codesandbox.io/p/sandbox/competent-cookies-4qmwh7?file=%2Fsrc%2Findex.ts%3A15%2C64

```typescript
import { create } from "@orama/orama";

(async () => {
  const orama = await create({
    schema: {
      foo: "string",
    },
    components: {
      tokenizer: {
        normalizationCache: new Map([["english:foo:Bar", "bar"]]),
        stemming: true,
      },
    },
  });
  console.log(orama.tokenizer.normalizationCache); // Empty map
})();

```
**Expected behavior**

Based on [documentation](https://docs.oramasearch.com/internals/components#tokenizer) regarding tokenizer configuration I would expect the provided map to be used when initializing a DB.

**Screenshots**

n/a

 - OS: `macOS 13.5.2`
 - Node version - 18.17.1
 - Browser - n/a - search done server side
 - Version - n/a - search done server side

**Smartphone (please complete the following information):**

n/a

**Additional context**

stearm:
Fixes #492.

Correctly create the tokenizer using the custom one passed via parameters.

Speaking about the linked issue, the provided example seems to be wrong:

```ts
const orama = await create({
    schema: {
      foo: "string",
    },
    components: {
      tokenizer: {
        normalizationCache: new Map([["english:foo:Bar", "bar"]]),
        stemming: true,
      },
    },
  });
```

According to the documentation, a custom tokenizer is made up of at least `language`, `normalizationCache`, and the `tokenize` function (see [Tokenizer](https://github.com/oramasearch/orama/blob/main/packages/orama/src/types.ts#L730) interface). For the time being, passing a custom `normalizationCache` is only possible if a custom tokenizer is used.

Passing a custom normalisation cache with default values without using a custom `tokenize` function could be harmful (or at the very least ineffective), because it would be supplied with specific keys determined by the language inside the native tokenizer.

### DIFF
--- a/packages/orama/src/methods/create.ts
+++ b/packages/orama/src/methods/create.ts
@@ -105,7 +105,7 @@ export async function create<
     id = await uniqueId()
   }
 
-  let tokenizer = components.tokenizer as Tokenizer
+  let tokenizer = components.tokenizer
   let index: TIndex | undefined = components.index
   let documentsStore: TDocumentStore | undefined = components.documentsStore
   let sorter: TSorter | undefined = components.sorter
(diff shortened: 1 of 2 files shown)


micheleriva (approved):
LGTM

PR #539, opened 2023-11-01
feat: export language with each stemmer

### LINKED ISSUE #510: Indexing of special characters
We have an issue searching for special characters, such as `æ`,`ø` and `å` (Danish letters).

Emojis also seems to cause problems.

```
"@orama/orama": "1.2.10",
"@orama/stemmers": "1.2.10"
```

rishi-raj-jain:
fix #510
/claim #510

### DIFF
--- a/packages/docs/pages/open-source/text-analysis/stemming.mdx
+++ b/packages/docs/pages/open-source/text-analysis/stemming.mdx
@@ -19,7 +19,7 @@ When stemming is enabled, Orama uses **the English language analyzer**, but we c
 
 ```javascript copy
 import { create } from '@orama/orama'
-import { stemmer } from '@orama/stemmers/italian'
+import { stemmer, language } from '@orama/stemmers/italian'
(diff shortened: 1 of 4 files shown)


rishi-raj-jain:
The implementation of `createTokenizer` function did accept stemmer as a whole, but was a exported function and not an object from which language could simply be extracted.

I've added another export to `language` which contains the correct language and can be used with the current props as is. 

Updated the example usage as well.

rishi-raj-jain:
@micheleriva

allevo:
Could you add a test to check if issue #510 is resolved?
Thanks

allevo:
Hi @rishi-raj-jain, could you add a test to check if issue #510 is resolved?
Thanks

rishi-raj-jain:
@allevo @micheleriva I've updated the tests to use the updated import.

micheleriva (approved):

PR #548, opened 2023-11-13
Adds basic telemetry plugin

micheleriva:


packages/plugin-telemetry/src/polyfills.ts
```diff
@@ -0,0 +1,14 @@
...
+    return
+  }
+
+  fetch(endpoint, {
+    method: 'POST',
+    body,
+    headers: {
+      'Content-Type': 'application/json'
+    }
+  })
```
allevo, line 13: We should capture the error so the nodejs process doesn't crash.

micheleriva: I'll just add a catch block here, we'll also need to discuss how to manage retries later

PR #553, opened 2023-11-19
chore: avoid recalculating lengths in some for-loops

aleattene:
To improve performance, I fixed some for-loops where the length was recalculated at each iteration, instead of calculating it just once before the loop itself.
Reviewer: @micheleriva


micheleriva (approved):
LGTM. Thanks!

PR #560, opened 2023-12-01
feat: enable prefix search for positive tolerance

### LINKED ISSUE #544: Tolerance disables prefix search
### Describe the bug

Usually e.g. `Chris` matches `Christopher`, but setting `tolerance` to 1 makes it not match anymore. I'm not sure whether this is intentional, but it's at least surprising.

### To Reproduce

```ts
      const db = await create({
        schema: {
          name: 'string',
        } as const,
        components: {
          tokenizer: {
            stemming: true,
            stopWords: englishStopwords,
          },
        },
      })

      await insert(db, { name: "Christopher" })
      const result = await search(db, { term: 'Chris', tolerance: 1 })
      // result should contain the document, but does not
```

### Expected behavior

Ideally I'd like to be able to search with tolerance _and_ allow matching prefixes.

### Environment Info

```bash
Orama 2.0.0-beta.4
```

### Affected areas

Search

### Additional context

If it's not intentional, I could look into fixing this.

rishi-raj-jain:
fix #544
/claim #544

## Solution Context

With tolerance: 0, the said case (of having prefix) would work as expected.

With increasing tolerance, I'd anyway expect the result to match the one it did for tolerance: 0 as the maximum distance has only increased.

This PR aims to implement the expectation/flow as above.

## Summary

- Add a `prefixSearch` optional option to search params
- By default `prefixSearch` is true
- Tweak the find function to accept the `prefixSearch`

## Testing

- Add tests for match and NO match when tolerance > 0, when  prefixSearch is `true` and `false` respectively


rishi-raj-jain:
@micheleriva 

https://github.com/oramasearch/orama/issues/544#issuecomment-1836851343 is what I tried to implement in this PR.

rishi-raj-jain:
What comes to my mind on top of this is to have a flag that'll explicitly look for a flag if tolerance is greater than 0 and decide to search for prefix. Wonder if this makes sense, and what should the default value of the flag be.

micheleriva:
@rishi-raj-jain, yes that makes a lot of sense to me. Could you please add some tests to verify that it's working correctly?

rishi-raj-jain:
Please assign the issue to me while I wrap this up? Thanks 🙏

micheleriva:
@rishi-raj-jain Since this issue is subject to a bounty, we prefer not to assign it to anyone - as it would discourage participants from applying their solution. I'm sure you can understand 🙏

rishi-raj-jain:
@micheleriva 

Added tests and did prop drilling.

rishi-raj-jain:
@micheleriva all the tests pass wohooo!

rishi-raj-jain:
@micheleriva 

## Summary of the changes I did:

- Add a `prefixSearch` optional option to search params
- By default `prefixSearch` is true
- Tweak the find function to accept the `prefixSearch`
- Add tests for match and NO match when tolerance > 0, when  prefixSearch is true and `false` respectively

## Tests pass!

rishi-raj-jain:
@micheleriva 

## Summary of the changes I did:

- Add a `prefixSearch` optional option to search params
- By default `prefixSearch` is true
- Tweak the find function to accept the `prefixSearch`
- Add tests for match and NO match when tolerance > 0, when  prefixSearch is true and `false` respectively

## Tests pass!

rishi-raj-jain:
@micheleriva 

I've added the tests, ready for your review.

rishi-raj-jain:
@micheleriva

I've added the tests, ready for your review.

With this contribution the DX is unmatched:

- every integration remains as is, the tolerance issue is fixed without any change from the user's end
- we update the core logic of find **without any hack** to create two searches: with tolerance 0 and tolerance 1 and then show the combination / common of these results
- we provide a flag to the user that they can use to opt out of tolerance search

I'm glad I could analyze all this, obtain your confirmation on my approach and make the PR with tests at the first go

rishi-raj-jain:
@micheleriva

I've added the tests, ready for your review.

With this contribution the DX is unmatched:

- every integration remains as is, the tolerance issue is fixed without any change from the user's end
- we update the core logic of find **without any hack** to create two searches: with tolerance 0 and tolerance 1 and then show the combination / common of these results
- we provide a flag to the user that they can use to opt out of tolerance search

I'm glad I could analyze all this, obtain your confirmation on my approach and make the PR with tests at the first go

rishi-raj-jain:
@micheleriva please take a look?

micheleriva:
Hi @rishi-raj-jain, there are multiple attempts for this bounty so please allow us to review all of them carefully before closing any

micheleriva:
Superseded by #580

rishi-raj-jain:
@micheleriva

How's the awarded fix any different from mine (that I came up with first within hours)?

https://github.com/oramasearch/orama/pull/560/files#diff-0a81a5033e2159840deb1d340175b3174049271f7744023924a53dc3e5087865R226

I've also added a flag that helps disable prefix search as we se agreed on https://github.com/oramasearch/orama/pull/560/#issuecomment-1837185755

SP321:
> @micheleriva
> 
> How's the awarded fix any different from mine (that I came up with first within hours of issue creation)?
> 
> https://github.com/oramasearch/orama/pull/560/files#diff-0a81a5033e2159840deb1d340175b3174049271f7744023924a53dc3e5087865R226
> 
> I've also added a flag that helps disable prefix search as we se agreed on #560
Your fix fails the testcase I wrote, Any branching in radix tree after finding the prefix is not traversed properly so it would only return a subset of all matches.

rishi-raj-jain:
Checking...

rishi-raj-jain:
Cool ✅

Well done!

PR #564, opened 2023-12-06
add `usePrefixWithTolerance`  param

### LINKED ISSUE #544: Tolerance disables prefix search

jkcs:
# This PR does the following: 
1. Adds the `usePrefixWithTolerance` option to allow executing the default prefix search when tolerance is not 0. 
2. Splits the `find` method in the `radix.ts` into default search and adds a method to merge the results.

close #544
/claim #544

# Request
please let me know first if this solution is feasible before I proceed with adding tests and other tasks.

### DIFF
--- a/packages/orama/src/components/index.ts
+++ b/packages/orama/src/components/index.ts
@@ -491,8 +491,8 @@ export async function search<T extends AnyOrama, ResultDocument = TypedDocument<
     throw createError('WRONG_SEARCH_PROPERTY_TYPE', prop)
   }
 
-  const { exact, tolerance } = context.params
-  const searchResult = radixFind(node, { term, exact, tolerance })
+  const { exact, tolerance, usePrefixWithTolerance } = context.params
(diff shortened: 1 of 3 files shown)


micheleriva:
Superseded by #580

PR #567, opened 2023-12-07
chore: use biome to format code

ematipico:
Hi peeps, how are you? 👋 

I am proposing an alternative to Prettier, Biome. https://biomejs.dev/formatter/

My idea is to slowly adopt the tool, and slowly adopt also linter (alongside eslint for the time being) and import sorting.

The PR only applies changes to `plugin-vitepress` as a starting point. I am happy to guide the migration to the rest of the code base.

While I was replacing it, I realized that prettier isn't used in the code base, so I changed the biome options to meet `ts-standard` - the library that Michele uses. However, there are still some differences due to some inconsistencies in the code.

Let me know what you guys think. If the change is welcomed, we could use this PR to decide the formatting style of the code base.

Thank you :)

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -1,0 +1,25 @@
+{
+	"$schema": "https://biomejs.dev/schemas/1.4.1/schema.json",
+	"formatter": {
+		"lineWidth": 120
+	},
+	"files": {
+		"include": ["packages/plugin-vitepress"],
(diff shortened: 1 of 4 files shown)


micheleriva:
That's fantastic, thank you @ematipico !

micheleriva (approved):

PR #602, opened 2024-01-09
Fix #601 - Skipping of keys in Object.prototype in search function

### LINKED ISSUE #601: Updating Object.prototype causes search to fail
### Describe the bug

Adding a new property to Object.prototype causes search function to fail

### To Reproduce

1 - Import necessary functions of Orama(insert, search, create)
2 - Add a property to object prototype
3 - Create an example schema 
4 - Insert some data into it
5 - Search

Here is an example:

```js
import { insert, search, create } from '@orama/orama';

Object.prototype.justTesting = () => {};

const schema = {
    name: 'string',
};

const orama = await create({
    schema,
});

await insert(orama, {
    name: 'John Doe',
});

const result = await search(orama, {
    term: 'John Doe',
});

console.log(result);
```

Code above throws an error saying "TypeError: searchResult[key] is not iterable" and after some debugging we can clearly see the issue:

```js
    const { exact , tolerance  } = context.params;
    const searchResult = radixFind(node, {
        term,
        exact,
        tolerance
    });
    const ids = new Set();
    console.log(searchResult)
    for(const key in searchResult){
        console.log(key, searchResult[key])
        for (const id of searchResult[key]){
            ids.add(id);
        }
    }
    return context.index.calculateResultScores(context, index, prop, term, Array.from(ids));
```

Here is the [code snippet](https://github.com/oramasearch/orama/blob/main/packages/orama/src/components/index.ts#L492C8-L492C8) from Orama's source code that I took which caused the issue. I added some console.logs to see what was the problem and second console.log prints out: "justTesting [Function (anonymous)]" and that causes searchResult[key] being not iterable.

### Expected behavior

Search function should properly return a value without any issues even if Object.prototype has changed.

### Environment Info

```bash
OS: Windows 10 Version 22H2 OS Build 19045.2486
Node: v20.10.0
Orama: 2.0.0-beta.12 (also tried it with 1.0.0-beta.16 and got the same result)
```

### Affected areas

Search

### Additional context

_No response_

SpongeBed81:
Fixes #601 

I also conducted some speed tests for the search function after this implementation, and the speed results were the same. I made sure that it had no effect on search speed.

I am not entirely sure that this PR won't create some inconsistencies in the codebase. Therefore, a check from the development team would be highly appreciated.

### DIFF
--- a/packages/orama/src/components/index.ts
+++ b/packages/orama/src/components/index.ts
@@ -490,6 +490,11 @@ export async function search<T extends AnyOrama, ResultDocument = TypedDocument<
   const ids = new Set<InternalDocumentID>()
 
   for (const key in searchResult) {
+    //if same key is in the Object.prototype, skip it
+    if(key in Object.prototype && Object.prototype[key] === searchResult[key]) {
+      continue;
+    }
+
     for (const id of searchResult[key]) {
       ids.add(id)
     }


SpongeBed81:
Noticed that I forgot the change the branch. Closing PR to open with a different branch.

PR #622, opened 2024-02-02
Fix typo in hybrid-search doc

afrideva:
Currently refers to undefined `searchVector` function


micheleriva (approved):
Awesome, thank you 🙏

Issue #641, opened 2024-02-21
Relevance parameters are not respected


valstu:
### Describe the bug

When you make a search with Orama and use relevance parameters those doesn't seem to have any effect on search results (or their scores).
```ts
await search(db, {
  term: "cat",
  relevance: {
        k: 2,
        b: 1.25,
})
```

The problem seems to be this exact line https://github.com/oramasearch/orama/blob/8377849700d62176c0eeeaf09c12637479e3c3fb/packages/orama/src/methods/search-fulltext.ts#L32

```ts
params.relevance = Object.assign(params.relevance ?? {}, defaultBM25Params)
```
`params.relevance` is always overwritten by `defaultBM25Params`. I think the `Object.assign` parameters should be flipped around so then we would overwrite defaults with custom parameters instead the other way.

So the solution would be?
```ts
params.relevance = Object.assign(defaultBM25Params, params.relevance ?? {})
```

### To Reproduce

- Create db
- Insert few documents
- Adjust relevance values
- Scores stay the same

I was able to find out that 

### Expected behavior

I would expect the scores to change

### Environment Info

```bash
OS: MacOS 14.1.2
Node: 20.11.0
Orama: 2.0.7
```

### Affected areas

Search

micheleriva:
Fixed with #643, thank you!

PR #643, opened 2024-02-21
Fix wrong order of params

valstu:
This PR fixes issue #641 where relevancy params where in wrong order in `Object.assign`


micheleriva (approved):
LGTM thanks!

PR #644, opened 2024-02-22
Fix to get threshold 0 working as documented

### LINKED ISSUE #642: Threshold 0 not working as documented
### Describe the bug

Threshold section on documentation mentions following:
> In this case, the threshold property is set to 0, which means that only the document containing the "slim fit" keywords will be returned. This applies to all the document properties; if a keyword is found in a property, and another keyword is found in a different property, the document will be returned.

Meaning that if threshold is set to 0 only the documents that have both keywords "slim" and "fit" will be returned. I noticed few cases where this was not true.

When using threshold 0, the search results sometimes included documents that didn't actually have both searched keywords but their score was higher than the ones that did have. This was probably due to the length difference of these documents, one of them was short and one of them was super long. Meaning the super long document got better score even if only one search term was found.

I tried to figure out why this happens and found out that in the comments [algorithms.ts](https://github.com/oramasearch/orama/blob/main/packages/orama/src/components/algorithms.ts#L60C3-L72C4) we are making an assumption that 
>   // Find the index of the last result with all keywords.
  // Note that since score is multipled by 1.5 any time the token is encountered in results it means
  // that tokenScores and tokenKeywordsCount should always have the same order.

So we are trusting that `tokenScores` and `tokenKeywordsCount` are similarly sorted. Then we find the index of last item index in `tokenKeywordsCount` where tokenKeywordCount is equal to `keywordsCount`

Then this index is used to slice the results array, which goes wrong because sometimes the tokenScores and tokenKeywordsCount are not in same order.

While investigating this I also encountered this issue #641 

### To Reproduce

I made an simple example here https://replit.com/@valstu/FunctionalSpiffyPolygon#index.js This example indexes two documents, one long and one short one. 

Run the index files to see the search results. Threshold is set to 0, one would expect the search to return the short file since it contains both of the keywords, as defined in documentation.

### Expected behavior

When threshold is set to 0 you would expect the search to return only documents that have all the keywords, how ever this is not the case everytime.

### Environment Info

```bash
OS: Replit
Node: 20.10.0
Orama: 2.0.7
```

### Affected areas

Search

### Additional context

_No response_

valstu:
This PR is a suggestion to fix #642. Instead of two sets, one for token counts and one for scores we combine them into a single set. Now when sorting the list it guarantees that they're in right order. Read further information and try the test case of a bug on Replit, mentioned in the issue #642

### DIFF
--- a/packages/orama/src/components/algorithms.ts
+++ b/packages/orama/src/components/algorithms.ts
@@ -12,8 +12,7 @@ export function prioritizeTokenScores(
     throw createError('INVALID_BOOST_VALUE')
   }
 
-  const tokenScoresMap = new Map<InternalDocumentID, number>()
-  const tokenKeywordsCountMap = new Map<InternalDocumentID, number>()
+  const tokenScoresMap = new Map<InternalDocumentID, [number, number]>()
 
   const mapsLength = arrays.length
   for (let i = 0; i < mapsLength; i++) {
@@ -23,14 +22,14 @@ export function prioritizeTokenScores(
     for (let j = 0; j < entriesLength; j++) {
       const [token, score] = arr[j]
       const boostScore = score * boost
-      const oldScore = tokenScoresMap.get(token)
+      const oldScore = tokenScoresMap.get(token)?.[0]
 
       if (oldScore !== undefined) {
-        tokenScoresMap.set(token, oldScore * 1.5 + boostScore)
-        tokenKeywordsCountMap.set(token, tokenKeywordsCountMap.get(token)! + 1)
+        tokenScoresMap.set(token, [(oldScore * 1.5 + boostScore), (tokenScoresMap?.get(token)?.[1] || 0) + 1])
+
       } else {
-        tokenScoresMap.set(token, boostScore)
-        tokenKeywordsCountMap.set(token, 1)
+        tokenScoresMap.set(token, [boostScore, 1])
+
       }
     }
   }
@@ -38,7 +37,7 @@ export function prioritizeTokenScores(
   const tokenScores: TokenScore[] = []
 
   for (const tokenScoreEntry of tokenScoresMap.entries()) {
-    tokenScores.push(tokenScoreEntry)
+    tokenScores.push([tokenScoreEntry[0], tokenScoreEntry[1][0]])
   }
 
   const results = tokenScores.sort((a, b) => b[1] - a[1])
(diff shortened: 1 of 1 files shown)


valstu:
Let me know what you think

micheleriva:
@allevo can you please take a look at this?

micheleriva (approved):
That's actually a neat fix. I like the approach

packages/orama/src/components/algorithms.ts
```diff
@@ -80,9 +89,14 @@ export function prioritizeTokenScores(
     lastTokenWithAllKeywords = 0
   }
 
+  const resultsWithIdAndScore: [number, number][] = [];
+  for (let i = 0; i < keywordsPerToken.length; i++) {
+    resultsWithIdAndScore.push([keywordsPerToken[i][0], keywordsPerToken[i][1]]);
+  }
+
```
allevo, line 96:
```suggestion
  const resultsWithIdAndScore: [number, number][] = new Array(keywordsPerToken.length);
  for (let i = 0; i < keywordsPerToken.length; i++) {
    resultsWithIdAndScore[i] = [keywordsPerToken[i][0], keywordsPerToken[i][1]];
  }
```

suggestion for better performance

packages/orama/src/components/algorithms.ts
```diff
@@ -80,9 +89,14 @@ export function prioritizeTokenScores(
     lastTokenWithAllKeywords = 0
   }
 
+  const resultsWithIdAndScore: [number, number][] = [];
+  for (let i = 0; i < keywordsPerToken.length; i++) {
+    resultsWithIdAndScore.push([keywordsPerToken[i][0], keywordsPerToken[i][1]]);
+  }
```
allevo, line 95:
```suggestion
  const resultsWithIdAndScore: [number, number][] = new Array(keywordsPerToken.length);
  for (let i = 0; i < keywordsPerToken.length; i++) {
    resultsWithIdAndScore[i] = [keywordsPerToken[i][0], keywordsPerToken[i][1]];

```

allevo:
I put a comment, but generally, it is ok for me!
Thanks!

valstu:
> I put a comment, but generally, it is ok for me! Thanks!

Added changes, thanks!

packages/orama/src/components/algorithms.ts
```diff
@@ -89,10 +89,10 @@ export function prioritizeTokenScores(
     lastTokenWithAllKeywords = 0
   }
 
-  const resultsWithIdAndScore: [number, number][] = [];
+  const resultsWithIdAndScore: [number, number][] = new Array(keywordsPerToken.length);
   for (let i = 0; i < keywordsPerToken.length; i++) {
-    resultsWithIdAndScore.push([keywordsPerToken[i][0], keywordsPerToken[i][1]]);
-  }
```
micheleriva: Missing closing bracket here :)

valstu: yeah, just  a sec :)

micheleriva:
Thank you @valstu , will release this in a few hours.

PR #694, opened 2024-04-10
perf(orama): facets and hybrid search

ilteoood:
With this PR I would like to improve the speed of the code related to facets retrieval and hybrid search.


micheleriva (approved):
Awesome! Thanks a lot!

PR #705, opened 2024-04-23
Add support for custom source dir on vitepress plugin

dbritto-dev:


dbritto-dev:
@micheleriva can we merge this too?

PR #710, opened 2024-04-29
remove recursion in avl trees

bortolottidev:
As suggested by title, I've converted lessThan and greaterThan (avl tree) using stack instead of recursion.

Couldn't find a way to achieve the same thing with rangeSearch without altering the order of the result


micheleriva (approved):

micheleriva:
Thank you!

PR #714, opened 2024-05-07
perf(orama): prefer flatMap

ilteoood:
As a perf improvement, use `flatMap` instead of `.map(...).flat`


micheleriva (approved):
Thank you so much!

PR #718, opened 2024-05-10
[Plugin Vitepress] fix case when item has no h1

SaraVieira:
Prevents plugin from breaking when a page doesn't have an h1

closes https://github.com/askorama/orama/issues/701


allevo (approved):
LGTM

PR #719, opened 2024-05-13
Fix docussarus plugin when there are no docs

SaraVieira:
- Skips generating search for docs when not present
- Switches component when no docs since the hooks are not usable anymore
- Adds a blog post even if no headers
- Runs biome

closes https://github.com/askorama/orama/issues/646


raiindev (approved):
LGTM

PR #737, opened 2024-06-17
Several optimizations

masylum:
I've been profiling orama, since it's taking 4s to ingest around 4k documents on my project and I would like to lower this down. I've noticed several things: It's ingesting a large amount of empty strings which is useless CPU time and also it's recalculating ids redundantly. This commit tries to address this two issues.

Additionally, I've noticed that due to the async APIs, the code is spending most of it's time waiting for "run microtasks". I have no idea if it would be possible to compile those away, because right now it makes the default implementation much worse (performance-wise) in order for people to be able to provide their storage solution.

Lastly, I've also noticed that providing the ID of the document, makes the ID be stored as part of the document properties. I thought it would only be to replace the default orama ID. I will fix this myself in userland by using `getDocumentProperties`, but perhaps is good to either change this default or document it.

### DIFF
--- a/packages/orama/src/components/index.ts
+++ b/packages/orama/src/components/index.ts
@@ -114,12 +114,10 @@ export interface Index extends AnyIndexStore {
 export async function insertDocumentScoreParameters(
   index: Index,
   prop: string,
-  id: DocumentID,
+  internalId: DocumentID,
   tokens: string[],
   docsCount: number
 ): Promise<void> {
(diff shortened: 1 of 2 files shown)


masylum:
hmm.. if I remove ids with `getDocumentProperties` it does not work. So perhaps it should be fixed.

micheleriva:
Hi there! I fear you'll need to regenerate the test snapshots. The PR looks good then!

Thank you so much

masylum:
how do I do that?

PR #738, opened 2024-06-17
Better normalization cache

masylum:
They key seems to be too specific. Specially by using the prop, which
basically makes it redudant to cache tokens that are found in different
props. The goal of that cache seems to be to trade memory for time, but
right now seems to be storing equal computations in different keys which
basically is inefficient. The only thing that the prop is needed for is
the `stemmerSkipProperties`.

### DIFF
--- a/packages/orama/src/components/tokenizer/index.ts
+++ b/packages/orama/src/components/tokenizer/index.ts
@@ -16,15 +16,13 @@ export interface DefaultTokenizer extends Tokenizer {
 }
 
 export function normalizeToken(this: DefaultTokenizer, prop: string, token: string): string {
-  const key = `${this.language}:${prop}:${token}`
-
-  if (this.normalizationCache.has(key)) {
-    return this.normalizationCache.get(key)!
+  if (this.normalizationCache.has(token)) {
+    return this.normalizationCache.get(token)!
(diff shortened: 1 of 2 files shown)


micheleriva:
Hi there! Thank you so much for your PR. Were you able to run the tests locally?

masylum:
hey, not really! I also found another positive side-effect of simplifying the normalization cache. Right now, these are always cache misses, which means that using the highlight plugin is pretty slow: https://github.com/askorama/orama/blob/7512ca936ffa1543a0e267ae9e9b3d4187be0bdd/packages/plugin-match-highlight/src/index.ts#L74

masylum:
I can't run tests locally, I get this when doing `npm -i`

> npm ERR! Unsupported URL Type "workspace:": workspace:*

masylum:
so it looks like the regression was introduced here: https://github.com/askorama/orama/pull/350/files

I'm currently working it around doing my own tokenizer, but it's not ideal

micheleriva:
> I can't run tests locally, I get this when doing npm install

You should use `pnpm`! `pnpm install` works just fine :)

micheleriva:
closed due to inactivity

PR #753, opened 2024-07-26
docs(plugin-docusaurus-v3):  detail for deploy type

innocces:
The deploy configuration in the initial example of the documentation is true. will cause the user to copy directly and cause the compilation to fail. Adding a configuration type enumeration and a link to the relevant type will help users use the deploy type more quickly.

image
image

The recommended initial configuration can be changed to default


micheleriva (approved):
Thank you!

PR #761, opened 2024-08-13
chore: use new searchbox

rjborba:
Adds the new searchbox.

It has some issues, but next release should fix most of them.

### DIFF
--- a/packages/docs/src/components/Search.jsx
+++ b/packages/docs/src/components/Search.jsx
@@ -1,6 +1,38 @@
 import { useEffect, useState } from 'react'
 import { OramaSearchBox, OramaSearchButton } from '@orama/react-components'
 
+function useCmdK(callback) {
+  const [isCmdKPressed, setIsCmdKPressed] = useState(false)
+
+  useEffect(() => {
+    const handleKeyDown = (event) => {
+      if ((event.metaKey || event.ctrlKey) && event.key === 'k') {
+        event.preventDefault()
+        setIsCmdKPressed(true)
+        if (callback && typeof callback === 'function') {
+          callback()
+        }
+      }
+    };
+
+    const handleKeyUp = (event) => {
+      if ((event.metaKey || event.ctrlKey) && event.key === 'k') {
+        setIsCmdKPressed(false);
+      }
+    };
+
+    window.addEventListener('keydown', handleKeyDown)
+    window.addEventListener('keyup', handleKeyUp)
+
+    return () => {
+      window.removeEventListener('keydown', handleKeyDown)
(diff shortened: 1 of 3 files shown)


packages/docs/astro.config.mjs
```diff
@@ -11,7 +11,7 @@ export default defineConfig({
   site: "https://docs.orama.com",
   vite: {
     ssr: {
-      noExternal: ['nanoid']
+      noExternal: ['nanoid', '@orama/react-components']
```
g-francesca, line 14: in memory of the hell behind this line.... 😄

packages/docs/src/components/Search.jsx
```diff
@@ -1,16 +1,14 @@
...
-import { SearchBox, SearchButton } from '@orama/searchbox'
-import '@orama/searchbox/dist/index.css'
+import { OramaSearchBox, defineCustomElements } from '@orama/react-components'
+import '@orama/react-components/dist/index.css'
 
-export const client = new OramaClient({
-  api_key: 'NKiqTJnwnKsQCdxN7RyOBJgeoW5hJ594',
-  endpoint: 'https://cloud.orama.run/v1/indexes/orama-docs-bzo330'
-})
+defineCustomElements()
```
g-francesca, line 6: - [ ] to be removed with the next release of `@orama/react-components`

packages/docs/src/components/Search.jsx
```diff
@@ -1,16 +1,14 @@
 import { useEffect, useState } from 'react'
-import { OramaClient } from '@oramacloud/client'
-import { SearchBox, SearchButton } from '@orama/searchbox'
-import '@orama/searchbox/dist/index.css'
+import { OramaSearchBox, defineCustomElements } from '@orama/react-components'
+import '@orama/react-components/dist/index.css'
```
g-francesca, line 3: style should be already bundled actually, if you remove this import I expect you to not have problems with style

packages/docs/src/components/Search.jsx
```diff
@@ -53,19 +51,51 @@ export function Search(props) {
...
-      eq: currentCategory
-    }
-  } : {}
+  const oramaWhere = currentCategory
+    ? {
+        category: {
+          eq: currentCategory
+        }
+      }
+    : {}
```
g-francesca, line 60: not sure, but maybe we don't need really need this if it's used for facets

- [ ] check if we need to add add this `where`

g-francesca: checked that we need it to filter between open source and cloud doc

PR #772, opened 2024-09-05
docs: add documentation for filters behavior on array properties

### LINKED ISSUE #767: Documentation for filters behavior on array properties
### Problem Description

Hello folks, I noticed a small detail missing in the documentation. The [filters documentation](https://docs.orama.com/open-source/usage/search/filters) explains which operators are available for each type, but it doesn't cover how filters behave on array properties. 
This behavior is well documented in the [tests](https://github.com/askorama/orama/blob/main/packages/orama/tests/array.test.ts), but I think it’s worth including a small section in the documentation.

### Proposed Solution

I would suggest adding something like the following to the [filters documentation](https://docs.orama.com/open-source/usage/search/filters) page, right before the "Enum operators" section.

## String[] | Number[] | Boolean[] operators

The available operators depend on the type (string, number, or boolean) as described in the previous sections.
A document matches if at least one of the array elements satisfies the filter condition.

```javascript copy
const db = await create({
  schema: {
    title: "string",
    tags: "string[]",
    editions: "number[]",
    limited: "boolean[]",
  }
});

await insertMultiple(db, [
  {title: "a", tags: ["foo", "bar"], editions: [1990, 2024], limited: [false, false]},
  {title: "b", tags: ["foo"], editions: [1942, 2024], limited: [false, true]},
  {title: "c", tags: ["bar"], editions: [2020], limited: [false]},
])

// Books with tag foo
await search(db, {where: {tags: "foo"}}); // returns  a, b

// Books tagged either as foo or bar
await search(db, {where: {tags: ["foo", "bar"]}}); // returns a, b, c

// Books with a 2024 edition
await search(db, {where: {editions: {eq: 2024}}}); // returns a, b

// Books with a limited edition
await search(db, {where: {limited: true}}); // returns b
```

If this makes sense to you, I’d be happy to open a PR.

### Alternatives

_No response_

### Additional Context

_No response_

simonecarriero:
Fixes #767 adding documentation for filters behavior on array properties

### DIFF
--- a/packages/docs/src/content/docs/open-source/usage/search/filters.mdx
+++ b/packages/docs/src/content/docs/open-source/usage/search/filters.mdx
@@ -113,6 +113,40 @@ const results = await search(db, {
 });
 ```
 
+## String[] | Number[] | Boolean[] operators
+
+The available operators depend on the type (string, number of boolean) as described in the previous sections.
+A document matches if at least one of the array elements matches the filter condition.
+
+```javascript copy
+const db = await create({
+  schema: {
+    title: "string",
+    tags: "string[]",
+    editions: "number[]",
+    limited: "boolean[]",
+  }
+});
+
+await insertMultiple(db, [
+  {title: "a", tags: ["foo", "bar"], editions: [1990, 2024], limited: [false, false]},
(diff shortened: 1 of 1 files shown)


micheleriva (approved):
Thanks!

PR #776, opened 2024-09-06
feat: generate embeddings on insert via Secure Proxy plugin

micheleriva:
With this PR, Orama OSS will be able to generate embeddings at insert-time via the Secure Proxy plugin.

```js
import { create, insert } from '@orama/orama'
import { pluginSecureProxy } from '@orama/plugin-secure-proxy'

const secureProxy = pluginSecureProxy({
  apiKey: 'xyz',
  embeddings: {
    defaultProperty: 'embeddings',
    model: 'orama/gte-small',
    onInsert: {
      generate: true, // Generate the embeddings at insert-time
      properties: ['name', 'data.age'],
      verbose: true
    },
  },
  chat: {
    model: 'openai/gpt-4o'
  }
})

const db = await create({
  schema: {
    name: 'string',
    data: {
      age: 'number'
    },
    embeddings: 'vector[384]'
  },
  plugins: [secureProxy]
})

await insert(db, { name: 'Michele', age: 29 })
// since "verbose" is set to `true`, it will log:
// Generating embeddings for properties name, data.age : "Michele. 29."
```

### DIFF
--- a/packages/plugin-secure-proxy/src/index.ts
+++ b/packages/plugin-secure-proxy/src/index.ts
@@ -8,20 +8,6 @@ export type SecureProxyExtra = {
 
 export type LLMModel = ChatModel
 
-// export type SecureProxyPluginOptions = {
-//   apiKey: string
-//   defaultProperty: string
-//   models: {
-//     embeddings: EmbeddingModel
-//     chat?: LLMModel
-//   }
-//   embeddingsConfig?: {
-//     generate: boolean
-//     properties: string[]
-//     verbose?: boolean
-//   }
-// }
-
 export type SecureProxyPluginOptions = {
   apiKey: string
   embeddings: {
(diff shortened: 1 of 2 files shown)


packages/plugin-secure-proxy/src/index.ts
```diff
@@ -32,6 +53,27 @@ export function pluginSecureProxy(pluginParams: SecureProxyPluginOptions): Orama
       proxy,
       pluginParams
     },
+
+    async beforeInsert<T extends TypedDocument<any>>(_db: AnyOrama, _id: string, params: PartialSchemaDeep<T>) {
+      if (!pluginParams.embeddings?.onInsert?.generate) {
```
allevo, line 58: embedding is not optional, right? so we can remove the `?` after `embeddings` (here and below)

allevo (commented):
I put a non-blocking comment: should `embeddings` be optional? it make sense?
From the config object, it seems mandatory.

packages/plugin-secure-proxy/src/index.ts
micheleriva: That's right, `embeddings` is not optional. Thanks!

PR #782, opened 2024-09-10
feat: adds new package (Orama Switch)

micheleriva:
With Orama Switch, you can use the same APIs to access either Orama Cloud or Orama OSS.

For instance, this is how you would interact with Orama Cloud:

```js
import { Switch } from '@orama/switch'
import { OramaClient } from '@oramacloud/client'

const client = new OramaClient({
  endpoint: '<Your Orama Cloud Endpoint>',
  api_key: '<Your Orama Cloud API Key>',
})

const orama = new Switch(client)

const results = await orama.search({
  term: 'noise cancelling headphones',
  where: {
    price: {
      lte: 99.99
    }
  }
})
```

And this is Orama OSS:

```js
import { Switch } from '@orama/switch'

const db = await create({
  schema: {
    productName: 'string',
    price: 'number'
  }
})

const orama = new Switch(client)

const results = await orama.search({
  term: 'noise cancelling headphones',
  where: {
    price: {
      lte: 99.99
    }
  }
})
```

### DIFF
--- a/packages/switch/README.md
+++ b/packages/switch/README.md
@@ -15,7 +15,7 @@ You can use the same APIs to access either Orama Cloud or Orama OSS.
 For instance, this is how you would interact with Orama Cloud:
 
 ```js
-import { switch } from '@orama/switch'
+import { Switch } from '@orama/switch'
 import { OramaClient } from '@oramacloud/client'
 
 const client = new OramaClient({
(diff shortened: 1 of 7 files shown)


packages/switch/README.md
```diff
@@ -0,0 +1,65 @@
...
+import { create } from '@orama/orama'
+
+const db = await create({
+  schema: {
+    productName: 'string',
+    price: 'number'
+  }
+})
+
+const orama = new Switch(client)
```
g-francesca, line 51: I think you meant to pass `db` here

Issue #766, opened 2024-08-22
Zero division when computing scores after removing all documents from an index.


kushuh:
### Describe the bug

I had a weird issues where all scores would suddendly switch to `NaN`. I found the culprit after some investigation:

https://github.com/askorama/orama/blob/980eb586ebaffbfc3879cf1d3d7c642e6e14c2eb/packages/orama/src/components/index.ts#L165

Basically, when removing the last document from an index, `(docsCount - 1)` becomes 0, which results in a `NaN` that screws up every following update.

I managed to fix the issue locally by changing the line to:

```ts
if (docsCount > 1) {
  index.avgFieldLength[prop] =
      (index.avgFieldLength[prop] * docsCount - index.fieldLengths[prop][internalId]!) / (docsCount - 1);
} else {
  index.avgFieldLength[prop] = undefined;
}
```

### To Reproduce

Just remove all documents from an index:

```ts
const id = await insert(db, document);
await remove(db, id);
```

### Expected behavior

If I add documents again to an emptied index, scores should compute normally.

### Environment Info

```bash
OS: Windows 11
Node: v22.3.0
Orama: 2.0.23
```

### Affected areas

Data Insertion, Search

### Additional context

Its not just when the index is empty. Even after adding back some documents, scores will remain NaN forever, as long as the index has been emptied once.

micheleriva:
Hi @kushuh, thank you so much for spotting this. Would you be up for opening a PR to fix this since you already fixed it locally? It would be immensely appreciated.

kushuh:
No problem, I'll open it by tonight

micheleriva:
Very appreciated. Thanks!

kushuh:
Hi @micheleriva, I tried to open the PR just now but I don't have permission to push a new branch

micheleriva:
Fixed in #789, released in Orama `v2.1.1`. Thanks again!

PR #789, opened 2024-09-14
Implement kushuh's change preventing div by zero in removeDocumentScoreParameters

AdventureBeard:
Implement @kushuh's proposed change in #766 , modifying removeDocumentScoreParameters to check for a divide by zero scenario before doing the calculation.

Sorry to snipe this change; just needed it badly, haha!


micheleriva:
Hi @AdventureBeard, thanks a lot for this PR! I fear that you should either do:

```js
index.avgFieldLength[prop] = 0;
```

or

```js
delete index.avgFieldLength[prop]
```

otherwise it'll fail type-checking as you're assigning `undefined` to a `number` property. Probably assigning to `0` is fine.

PR #796, opened 2024-09-26
Fix project build issues

matijagaspar:
* Changed node version in .nvmrc to 20 (18 fails the build)
* Removed conflicting configuration for swc: either jsc or env (removed env)
* fixed ts confings so ts builds pass
* fixed a test run script of plugin-astro (--loader no longer supported)
* plugin-docusaurs-v3 fixed postbuild.sh to not fail if dist directory is missing
* fixed so `pnpm run lint` passes


micheleriva (approved):
LGTM!

PR #808, opened 2024-10-02
Improvements

allevo:
Main changes:
- change IIndex interface (breaking change)

Other changes:
- remove context search object
- better-shared code for fulltext/vector/hybrid searches
- remove slovenian stemmer (we never have it) & fix russian stemmer (that was called russian for some strange reason)


micheleriva:
@allevo can we merge this?

PR #817, opened 2024-10-11
Add @orama/plugin-pt15

allevo:
Implement different index algorithm.


packages/plugin-pt15/src/algorithm.ts
```diff
@@ -0,0 +1,214 @@
+import { AnyIndexStore, AnyOrama, SearchableType, Tokenizer, VectorIndex } from "@orama/orama"
+import { avl, bkd, flat, bool } from '@orama/orama/trees'
+import {
+  getVectorSize, internalDocumentIDStore, isVectorType } from '@orama/orama/components'
+
+type InternalDocumentID = internalDocumentIDStore.InternalDocumentID;
+
+export type TreeType = 'AVL' | 'Radix' | 'Bool' | 'Flat' | 'BKD';
```
micheleriva, line 8: We should export these from the `@orama/orama` package

packages/plugin-pt15/src/algorithm.ts
```diff
@@ -0,0 +1,214 @@
...
+export type TTree<T = TreeType, N = unknown> = {
```
micheleriva, line 9: This one too

packages/plugin-pt15/src/algorithm.ts
```diff
@@ -0,0 +1,214 @@
...
+    type: T;
+    node: N;
+    isArray: boolean;
+};
+export type Tree =
```
micheleriva, line 14: This one too

packages/plugin-pt15/src/algorithm.ts
```diff
@@ -0,0 +1,214 @@
...
+  searchableProperties: string[]
+  searchablePropertiesWithTypes: Record<string, SearchableType>
+}
+
+function create_obj() {
+  // object with empty prototype to cheap objects
+  return Object.create(null)
+}
+
+export function recursiveCreate<T extends AnyOrama>(indexDatastore: PT15IndexStore, schema: T['schema'], prefix: string) {
```
micheleriva, line 56: Is recursion well supported in that case? Should we opt for an iterative approach for slightly larger datasets (50k elements)?

packages/plugin-pt15/src/algorithm.ts
```diff
@@ -0,0 +1,214 @@
...
+
+    if (isVectorType(type)) {
+      indexDatastore.searchableProperties.push(path)
+      indexDatastore.searchablePropertiesWithTypes[path] = type
+      indexDatastore.vectorIndexes[path] = {
+        size: getVectorSize(type),
+        vectors: {}
+      }
+    } else {
+      const isArray = /\[/.test(type as string)
```
micheleriva, line 74: we may want to use `.endsWith("[]")` or even `type[type.length-1] == ']'` for performance reasons

packages/plugin-pt15/src/algorithm.ts
```diff
@@ -0,0 +1,214 @@
...
+    }
+  }
+}
+
+export function insertString(
+  value: string,
+  positionsStorage: PositionsStorage,
+  prop: string,
+  internalId: InternalDocumentID,
+  language: string | undefined,
```
micheleriva, line 126: I think I started exposing a `Optional` type which is basically `type Optional<T = unknown> = T | undefined`

micheleriva (approved):
minor comments, but lgtm!

packages/plugin-pt15/src/algorithm.ts
allevo: This is not an intensive path: this function is invoked only during the orama creation

PR #828, opened 2024-10-18
fixed searchString tokensLength error

Vvanlaar:


allevo:
Hi!
Thank you for your contribution.
Could you explain better which error you have?

Vvanlaar:
I'm getting this error:
Cannot read properties of undefined (reading 'tokensLength')
I think it should read stats from prop since the function parameter is only the prop object with a stats property.

allevo:
Can you write a code example that generates the error?

micheleriva:
Closing due to inactivity.

PR #853, opened 2024-12-09
Fix: async processing for hooks

niltonheck:


micheleriva (approved):
This will be subject to more fixes late but as for now, lgtm

PR #870, opened 2025-01-17
fix: Astro 5 Compatibility

### LINKED ISSUE #862: The Astro plugin is not compatible with Astro 5
### Describe the bug

Astro 5 introduces breaking changes that affect this plugin. According to astro documentation:

"Astro v5.0 updates the shape of IntegrationRouteData.distURL to be undefined or an array of URLs. This fixes a previous error because a route can generate multiple files on disk, especially when using dynamic routes such as [slug] or [...slug]"

This change made to the `distUrl` used by the plugin `prepareOramaDb` causes the plugin to throw the following error before completing:
```
Cannot read properties of undefined (reading 'endsWith')
Location:
  /Users/.../node_modules/@orama/plugin-astro/dist/index.js:40:70
```

### To Reproduce

1. Create an astro 5 project
2. Install and configure the orama plugin
3. Build the project

### Expected behavior

The porject builds and the orama db is created

### Environment Info

```bash
OS: Macos 14.7.1
Node: v18.20.1
```

### Affected areas

Initialization

### Additional context

After reviewing the code looks like an easy fix that can be done backwards compatible so it won't require a major relase of the plugin

awhitford:
Astro v5.0 updates the shape of `IntegrationRouteData.distURL` to be undefined or an array of URLs.

Fixes #862

### DIFF
--- a/packages/plugin-astro/src/index.ts
+++ b/packages/plugin-astro/src/index.ts
@@ -58,13 +58,15 @@ async function prepareOramaDb(
 
   // All routes are in the same folder, we can use the first one to get the basePath
   const basePath = dir.pathname.slice(isWindows ? 1 : 0)
+  // Create a dist urls
+  const distUrls = routes.flatMap((r) => r.distURL)
   const pathsToBeIndexed = pages
     .filter(({ pathname }) => dbConfig.pathMatcher.test(pathname))
     .map(({ pathname }) => {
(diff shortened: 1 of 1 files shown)


micheleriva:
Hey @awhitford thanks for your contrib!

micheleriva (approved):

PR #885, opened 2025-02-11
chore: Updated Astro peer dependency to version 5

awhitford:
Also updated Node Engine to be at least 20 because Node 18 does not work.


micheleriva:
@awhitford the CI is not passing :(

micheleriva:
Closing due to inactivity

PR #888, opened 2025-02-12
Fix: #850 add threshold logic back

### LINKED ISSUE #850: Threshold not working at all
### Describe the bug

Threshold is not working at all, it returns the same results if it is equal to 1 or if it is equal to 0.

We can use the exact example given in the docs to reproduce the bug.

### To Reproduce

```ts
import { create, insert, search } from "@orama/orama";

const db = create({
  schema: {
    title: "string",
  },
});

insert(db, { title: "Blue t-shirt, slim fit" });
insert(db, { title: "Blue t-shirt, regular fit" });
insert(db, { title: "Red t-shirt, slim fit" });
insert(db, { title: "Red t-shirt, oversize fit" });

const defaultThreshold = search(db, {
  term: "regular fit",
}); // 4 hits - OK
const zeroThreshold = search(db, {
  term: "regular fit",
  threshold: 0,
}); // Also 4 hits - NOT OK

```

### Expected behavior

The threshold should work as described in the docs.

### Environment Info

```bash
Orama: 3.0.2
Tried with BM25, QPS, and PT15, and it's not working on any of them.
```

### Affected areas

Search

### Additional context

_No response_

matijagaspar:
Fix #850

### DIFF
--- a/packages/orama/src/components/index.ts
+++ b/packages/orama/src/components/index.ts
@@ -387,16 +387,16 @@ export function calculateResultScores(
   bm25Relevance: Required<BM25Params>,
   resultsMap: Map<number, number>,
   boostPerProperty: number,
-  whereFiltersIDs: Set<InternalDocumentID> | undefined
+  whereFiltersIDs: Set<InternalDocumentID> | undefined,
+  keywordMatchesMap: Map<InternalDocumentID, Map<string, number>>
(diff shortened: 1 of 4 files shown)


micheleriva (approved):
Thank you so much Matija 🙏

PR #894, opened 2025-02-24
setup(publish): support JSR

AugustinMauroy:
## Info

- `tokenizers` use slow type because build script didn't output type def for `tokenizer.js`
-  we only publish orama package and because the rest is plugin and sub framework isn't on JSR 

## Quesiton

- Should I need to introduce a new GA that publish package (it's will allow you to have provenance)


AugustinMauroy:
bump @micheleriva

micheleriva:
Hey @AugustinMauroy sorry for the delay. I'll try to publish this as a RC on JSR to see how it goes and will give you a feedback 🙏

AugustinMauroy:
bump @micheleriva

PR #902, opened 2025-03-03
Fix examples in README.md

ggodlewski:
Similarity of best match is 0.799 so I decreased threshold to 0.75


micheleriva (approved):
Thanks!

Issue #475, opened 2023-09-07
Positions are duplicated with `afterInsert` in `@orama/plugin-match-highlight` 


benjaminleonard:
**Describe the bug**
Positions are duplicated with `afterInsert` in `@orama/plugin-match-highlight` but only in the imported function. I copied it into my repo to test and it fixes the issue. Could be an issue with CJS vs ESM? In this instance it is using the converted function as below:

```
async function afterInsert(...args) {
    if (!_esmAfterInsert) {
        const imported = await import('./index.js');
        _esmAfterInsert = imported.afterInsert;
    }
    return _esmAfterInsert.apply(this, args);
}
```

**To Reproduce**
Difficult to reproduce because only some positions appear multiple times.

**Expected behavior**
That positions only appear once, but instead some positions appear twice — as if `afterInsert` is running more than once.

See correct version with inline `highlightAfterInsert`
```json
        "ip": [
          {
            "start": 14,
            "length": 2
          },
          {
            "start": 73,
            "length": 2
          },
          {
            "start": 118,
            "length": 2
          },
          {
            "start": 336,
            "length": 2
          },
          {
            "start": 441,
            "length": 2
          },
          {
            "start": 474,
            "length": 2
          },
          {
            "start": 581,
            "length": 2
          },
          {
            "start": 639,
            "length": 2
          }
        ],
```

Duplicated with imported `highlightAfterInsert`
```json
"ip": [
          {
            "start": 14,
            "length": 2
          },
          {
            "start": 73,
            "length": 2
          },
          {
            "start": 118,
            "length": 2
          },
          {
            "start": 336,
            "length": 2
          },
          {
            "start": 441,
            "length": 2
          },
          {
            "start": 474,
            "length": 2
          },
          {
            "start": 581,
            "length": 2
          },
          {
            "start": 639,
            "length": 2
          },
          {
            "start": 73,
            "length": 2
          },
          {
            "start": 118,
            "length": 2
          },
          {
            "start": 336,
            "length": 2
          },
          {
            "start": 441,
            "length": 2
          },
          {
            "start": 474,
            "length": 2
          },
          {
            "start": 581,
            "length": 2
          },
          {
            "start": 639,
            "length": 2
          },
          {
            "start": 14,
            "length": 2
          },
          {
            "start": 73,
            "length": 2
          },
          {
            "start": 118,
            "length": 2
          },
          {
            "start": 336,
            "length": 2
          },
          {
            "start": 441,
            "length": 2
          },
[...]

PR #905, opened 2025-03-08
Fix bundling and type issues in match-highlight plugin

fturmel:
I understand that you have been deprecating the plugin, but I'd like to propose that you consider publishing these types and bundling fixes so that it's at least usable with v3.

- The docs for it are still available without deprecation warnings: https://docs.orama.com/open-source/plugins/plugin-match-highlight

- The new highlight package doesn't have functional parity (can't highlight full words on partial matches, can't highlight searches with typo tolerance) https://github.com/oramasearch/highlight/issues/13
 
- Maybe it was a mistake, but there _is_ a published v3 version of the plugin on npm but it's not working properly with Orama v3 (very confusing). Again, I know it's deprecated but to me it's still giving the impression that it should be working with Orama v3 and will stop working when the next major version is released. https://www.npmjs.com/package/@orama/plugin-match-highlight

I am vendoring the plugin in my project now, which has fixed the duplicate insert bug mentioned here https://github.com/oramasearch/orama/issues/475, so I'm hoping that using `tsup` to bundle the plugin like the other packages will get rid of that strange behavior. I copied over the tsconfig.json and tsup.lib.js files from one of the other packages, updated the package.json scripts and export, etc. I think it should be sufficient?

Figured I'd at least open a PR to contribute this back in case I'm not the only one in this situation and struggling with the plugin. If you're not interested in merging or reviewing, at least the code will be out there for others to find and vendor as well.


micheleriva (approved):
Hey @fturmel , thank you so much for this PR. Although we do not officially support this plugin anymore, I'm happy to get any PR merged and released!

fturmel:
@micheleriva Appreciate the quick review and merge!

PR #927, opened 2025-04-26
Node 20 is the new minimum

awhitford:
Node 16 was EOL as of September 11, 2023, and Node 18 is EOL as of April 30, 2025, so might as well just make 20 the new minimum.

See [Node.js Releases](https://nodejs.org/en/about/previous-releases) for more information.


micheleriva (approved):
Agreed. Thanks

PR #928, opened 2025-04-26
Upgraded pnpm to 10.9.0 (from 8.15.4)

awhitford:
It is great to see `packageManager` explicitly declared in `package.json`, but there is an opportunity to upgrade it to the latest _stable_ version with the latest security updates.


micheleriva (approved):
Thanks!

PR #929, opened 2025-04-28
Docs - Starlight Refactoring

awhitford:
In preparation for a Starlight upgrade, I discovered that there were several opportunities to simplify the custom components and avoid reimplementing built-in component logic so that long term maintenance is easier, and it benefits from future enhancements.
* Removed the `Hero` component because it did not seem to deviate from the built-in component.
* Revised the `Footer` component to use the built-in component.
* Replaced the custom `Sidebar` component with a version that leverages the built-in component, and was able to remove the redundant `SidebarSubList` component.
* Removed the `Header` component in favor of simply overriding the `Search` component.  (This maintains the _Mobile_ menu.)
* Revised the `TableOfContents` component to use the built-in component.

There may be some slight cosmetic differences, but hopefully you will agree that this is easier to maintain and sets the project up to benefit from future enhancements, such as the `SidebarPersister`.


awhitford:
@micheleriva Please let me know if you have any questions or concerns about this PR.  Once merged, I will revisit the Astro and Starlight upgrade, but this will make things easier.

Saif-Shines:
I'm imaging this plugin will:

- Allow my to declare configuration about Orama
- Key in credentials from Orama Cloud
- The Search, Header or necessary UI components are automatically overriden.

Is that right? @awhitford ?

awhitford:
@Saif-Shines This PR simply includes some refactoring of how Starlight is used by Orama's Docs.  This PR is helpful because it will then enable me to upgrade to a newer version of Astro and Starlight for Orama Docs.

This does not impact Orama's Astro Plugin.  However, I would like to make changes to that eventually -- but this PR was meant to be a step in that direction.

Still patiently waiting after a month.

micheleriva:
We're moving to Fumadocs at https://github.com/oramasearch/docs

PR #940, opened 2025-05-18
deps: Upgraded tap test framework to 21.1.0

awhitford:
- Added a top-level `devDependency` declaration for `tap` version 21.1.0 (latest).
- Previously, there was a mix of `tap` versions being used:  mostly 18.6.1, but sometimes 18.7.1 or 21.0.1.
- Removed unnecessary `devDependency` declarations in sub packages.
- Removed legacy `@types/tap` dependency declarations.
- Also upgraded `tap-mocha-reporter` to 5.0.4, from 5.0.3.

Note that `packages/orama/src/package.json` was deleted -- that seems to be a side effect the build or test process, but the latest version doesn't seem to add any value.

Some minor json reformatting is also a result of the build/test process -- likely biome.


micheleriva:
@awhitford can you please fix conflicts?

micheleriva:
Closing due to inactivity

Issue #866, opened 2024-12-28
Search with `exact: true` doesn't work as intended


zeroliu:
### Describe the bug

When I search with `exact: true`, documents with partially matching terms are also returned.

### To Reproduce

https://codesandbox.io/p/sandbox/lvy7mc

const db = create({
  schema: {
    path: "string",
    title: "string",
  },
});

insert(db, { path: "First Note.md", title: "First Note" });
insert(db, { path: "Second Note.md", title: "Second Note" });

const noExact = search(db, {
  term: "first",
  properties: ["path"],
}); // 1 result OK
const withExact = search(db, {
  term: "first",
  properties: ["path"],
  exact: true,
}); // 1 result NOT OK

console.log("noExact path", noExact);
console.log("withExact path", withExact);

### Expected behavior

Calling `search` with `exact: true` should only return documents with terms that exactly matches.

### Environment Info

```bash
@orama/orama 3.0.4
```

### Affected areas

Search

logancyang:
Hello? Anybody working on this project?

micheleriva:
Hi @logancyang and @zeroliu, we are currently experiencing a temporary slowdown in managing this activity. We will keep you updated.

micheleriva:
@matijagaspar is this fixed with your last PR?

joybro:
I'm experiencing this issue with the latest version (3.1.1).

Mobarak786:
I understand the problem. It happens because  the adject match in search matching with "First" But not with "Note.md"  And this is the behavior of "orama"  as it tokenise the name when an space occours. Thats why you are getting the error. 

@zeroliu  Could you assign to me. I can fix it.
/assign  #866

zeroliu:
Thanks for looking into it @Mobarak786 . I don't have the permission to assign the ticket though. Maybe a repo admin can help

micheleriva:
Hi @Mobarak786 , @zeroliu , thanks for your help! You don't need to get the issue assigned, just send a PR and if it gets merged, we will pay the bounty. Thanks again!

Mobarak786:
Does the bounty still available.

Sambit003:
@micheleriva is this issue still open? if. then I will start working on it

DecodersLord:
@micheleriva I believe what @Mobarak786 pointed out is correct. 

Can you advise if the output should only include if the document had "first.md" in the user's test case?

Omar8345:
[...]

PR #942, opened 2025-05-24
fix: ensure exact:true matches entire property value, not tokens

luffy-orf:
**PR Description**

- **Fixes exact search behavior:**  
  - Ensures that `exact: true` matches only when the *entire property value* (not just a token/word) matches the search term, case-insensitively.
  - This prevents partial/token matches (e.g., `"First Note.md"` will not match `"first"` with `exact: true`).

- **Distinguishes between property-value and token-level exactness:**  
  - `exact: true` → matches the whole property value.
  - `exactToken: true` → matches individual tokens/words exactly.

- **Adds/clarifies documentation and code comments:**  
  - Clearly explains the difference between `exact` and `exactToken` in both code and types.

- **How this solves the issue:**  
  - Users searching with `exact: true` now get results only when the property value is an exact match, as expected.
  - No more false positives from partial/token matches.

Solves - #866 
/claim #866

### DIFF
--- a/packages/orama/src/methods/search-fulltext.ts
+++ b/packages/orama/src/methods/search-fulltext.ts
@@ -1,7 +1,7 @@
 import { getFacets } from '../components/facets.js'
 import { getGroups } from '../components/groups.js'
 import { runAfterSearch, runBeforeSearch } from '../components/hooks.js'
-import { getInternalDocumentId } from '../components/internal-document-id-store.js'
+import { getInternalDocumentId, InternalDocumentID } from '../components/internal-document-id-store.js'
 import { Language } from '../components/tokenizer/languages.js'
 import { createError } from '../errors.js'
 import type {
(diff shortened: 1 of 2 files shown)


luffy-orf:
@zeroliu @micheleriva , 
Could you please review the PR and let me know if any changes are needed?

Also, just wanted to mention that this is my first time contributing here — happy to be part of the project!

Thanks!

zeroliu:
Do you know what the best way is to test these changes on my end? Can some unit tests be included? I only used Orama as a user and not very familiar with the process either

vachmara:
Hi @luffy-orf,

I wanted to clarify the context between PR #941 (my submission) and PR #942.
PR #942 appears to be a direct copy of my code from #941, with only a comment added, so it doesn’t address the full scope, including test coverage and refactoring, that I provided.

Regarding the renaming of `exact` to `exactToken` @zeroliu :
I introduced this breaking change to **avoid removing a feature that existing Orama users may rely on**. By making the distinction explicit, we can support both behaviors more clearly. All related tests have been updated to reflect this.

I do recognize that introducing a breaking change isn’t ideal for everyone, and I’m open to suggestions, such as creating a new search option like `strictlyExact` to maintain full backward compatibility if that’s preferred.

If you’d like to test the changes locally, I recommend:
```bash
git clone https://github.com/vachmara/orama.git
cd packages/orama
pnpm i 
pnpm run test
```
You can experiment further in the [`search.test.ts`](https://github.com/vachmara/orama/blob/main/packages/orama/tests/search.test.ts) file.

Happy to iterate on this based on the maintainers’ and community’s feedback!

Thanks!

PR #952, opened 2025-06-22
doc: update plugin-secure-proxy.mdx

rexwangcc:
It seems it's either a typo or an outdated code block?


micheleriva (approved):
Nice catch, thanks!

PR #963, opened 2025-08-19
Add esbuild bundle script

### LINKED ISSUE #961: Circular dependencies / require cycle
### Describe the bug

Surprised this has not been reported yet. I've used Orama previously with Vite as a bundler without issues.

I'm now setting up an Expo web/native project, and see these warnings:

```
 WARN  Require cycle: ../../../node_modules/@orama/orama/dist/esm/errors.js -> ../../../node_modules/@orama/orama/dist/esm/utils.js -> ../../../node_modules/@orama/orama/dist/esm/errors.js

Require cycles are allowed, but can result in uninitialized values. Consider refactoring to remove the need for a cycle.
 WARN  Require cycle: ../../../node_modules/@orama/orama/dist/esm/methods/search.js -> ../../../node_modules/@orama/orama/dist/esm/methods/search-fulltext.js -> ../../../node_modules/@orama/orama/dist/esm/methods/search.js

Require cycles are allowed, but can result in uninitialized values. Consider refactoring to remove the need for a cycle.
 WARN  Require cycle: ../../../node_modules/@orama/orama/dist/esm/methods/search.js -> ../../../node_modules/@orama/orama/dist/esm/methods/search-hybrid.js -> ../../../node_modules/@orama/orama/dist/esm/methods/search.js
```

These are not benign, it actually resulted in a `fetchDocuments is not a function` runtime error when calling the `search` function.

At the moment, the only workaround I can think of would be for users to bundle the entirety of the Orama source into a single file via esbuild or similar.

### To Reproduce

1. `npx create-expo-app@latest test`
2. `cd test && npm i @orama/orama`
3. `npm run ios`

### Expected behavior

No require cycles – shared dependencies should be extracted to a third file to break the circularity.

### Environment Info

```bash
N/A
```

### Affected areas

Search

### Additional context

_No response_

isaachinman:
Fixes #961

### DIFF
--- a/packages/orama/package.json
+++ b/packages/orama/package.json
@@ -16,6 +16,10 @@
         "types": "./dist/browser/index.d.ts",
         "default": "./dist/browser/index.js"
       },
+      "bundle": {
+        "types": "./dist/esm/index.d.ts",
+        "default": "./dist/bundle/index.js"
+      },
       "import": {
         "types": "./dist/esm/index.d.ts",
         "default": "./dist/esm/index.js"
@@ -33,6 +37,10 @@
       "browser": {
         "types": "./dist/browser/internals.d.ts",
         "default": "./dist/browser/internals.js"
+      },
+      "bundle": {
+        "types": "./dist/esm/internals.d.ts",
+        "default": "./dist/bundle/internals.js"
       },
       "import": {
         "types": "./dist/esm/internals.d.ts",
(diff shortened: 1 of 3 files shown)


thatjuan:
Superseded by #1026, now merged. Rather than shipping a bundled artifact, #1026 fixes this at the source: it extracts `fetchDocuments`/`fetchDocumentsWithDistinct` out of `methods/search.ts` into their own module so the per-mode search modules no longer import `search.ts` back — removing the circular dependency entirely. That fixes it for every bundler (no `bundle` export condition or esbuild postbuild for consumers to opt into) and keeps the build bundler-free, which avoids the ESM↔CJS concern raised in #961. Thanks for the clear repro and the example PR — they made the root cause easy to pin down. Closing in favor of #1026.

PR #968, opened 2025-09-04
fix: exact search with exact: true

### LINKED ISSUE #866: Search with `exact: true` doesn't work as intended

melmathari:
## 🐛 Bug Fix

Fixes #866 - exact search with `exact: true` now matches entire property values instead of individual tokens.

When using `exact: true`, users expected the search to match the entire property value, but Orama was only performing exact token matching. This caused confusing and unexpected behavior:

**Example of the bug:**
```javascript
const db = create({ schema: { path: "string" } });
insert(db, { path: "First Note.md" });

// This would incorrectly return a match
search(db, { term: "first", exact: true, properties: ["path"] });
```

**Expected behavior:** No match (because `"first" !== "First Note.md"`)  
**Actual behavior:** ❌ Match found (because token `"first"` exists in the tokenized property)

### ✅ Solution

Implemented property-level exact matching with post-filtering approach:

1. **Maintains performance**: Uses existing tokenized search to find candidates
2. **Adds exact filtering**: When `exact: true`, filters results to only include documents where at least one searched property exactly matches the search term (case-insensitive)
3. **Backward compatible**: Doesn't break existing functionality or API

**Technical implementation:**
- Added filtering logic in `search-fulltext.ts` after the initial search
- Uses `getNested()` utility to access property values from documents
- Performs case-insensitive exact comparison of entire property values
- Works with single properties, multiple properties, and property arrays

### 🧪 Testing

Added comprehensive test coverage:
- **New test case**: `"should match entire property value when exact is true"`
- **Standalone verification**: Created `test-exact-fix.mjs` demonstrating the fix
- **Validation results**:
  ```
  ✅ EXACT SEARCH FIX IS WORKING!
  - Searching for 'first' with exact:true only matches documents where path is exactly 'first'
  - Searching for 'First Note.md' with exact:true matches the correct document
  ```

### 📊 Before vs After

| Search | Document | exact: true | Before | After |
|--------|----------|-------------|---------|--------|
| `"first"` | `{ path: "First Note.md" }` | ✅ | ❌ Match | ✅ No match |
| `"first"` | `{ path: "first" }` | ✅ | ✅ Match | ✅ Match |
| `"First Note.md"` | `{ path: "First Note.md" }` | ✅ | ✅ Match | ✅ Match |

### 🔄 Breaking Changes

**None** - This change fixes the behavior to match user expectations. The `exact: true` parameter already existed; we just corrected its implementation.

### 📁 Files Changed

- `packages/orama/src/methods/search-fulltext.ts` - Added property-level exact matching logic
- `packages/orama/tests/search.test.ts` - Added comprehensive test coverage
- `packages/orama/tests/test-exact-fix.mjs` - Standalone demonstration script

### 🎯 Resolves

- Closes #866
- Addresses user confusion about `exact: true` behavior
- Aligns implementation with intuitive user expectations

---

### 🧪 How to Test This Fix
```bash
# Build the project first
npm run build

# Run our standalone test
node tests/test-exact-fix.mjs
```

**Bounty Claim:** /claim #866

This fix ensures that `exact: true` now works as users intuitively expect - matching entire property values rather than just tokens within them.

### DIFF
--- a/packages/orama/src/methods/search-fulltext.ts
+++ b/packages/orama/src/methods/search-fulltext.ts
@@ -15,7 +15,7 @@ import type {
   TokenScore,
   TypedDocument
 } from '../types.js'
-import { getNanosecondsTime, removeVectorsFromHits, sortTokenScorePredicate } from '../utils.js'
+import { getNanosecondsTime, getNested, removeVectorsFromHits, sortTokenScorePredicate } from '../utils.js'
 import { count } from './docs.js'
 import { fetchDocuments, fetchDocumentsWithDistinct } from './search.js'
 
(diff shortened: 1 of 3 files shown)


melmathari:
@micheleriva At your convenience, appreciate you taking the time.

melmathari:
@micheleriva for your reference. Appreciate if this can be reviewed.

PR #980, opened 2025-09-14
Update stopwords docs URL

szepeviktor:
The current one is https://docs.orama.com/docs/orama-js/text-analysis/stop-words


micheleriva (approved):
Thank you so much

PR #983, opened 2025-09-26
Plugin for the PgVector

saquib34:
This pull request introduces a new `@orama/plugin-pgvector` package, enabling Orama to use PostgreSQL with the pgvector extension as a backend for persistent, memory-efficient vector search and storage. The implementation offloads document storage and vector operations to PostgreSQL, supports lazy loading and memory limits, and provides a test script for quick validation.

The most important changes are:

**New Plugin Implementation:**

* Added the `@orama/plugin-pgvector` package with a `package.json` defining dependencies on `pg`, `pgvector`, TensorFlow libraries, and Orama, and including build/test scripts.
* Implemented the main plugin logic in `src/index.ts`, which:
  - Hooks into Orama's lifecycle to store documents in PostgreSQL, remove them from memory, and perform vector searches directly in the database.
  - Supports lazy loading, memory limits, and a vector-search-only mode for efficiency.
  - Manages document caching and memory usage with an LRU strategy.

**Database Management and Vector Operations:**

* Added `src/pgvector-manager.ts` to handle:
  - Table creation, schema adaptation, and vector column/index management in PostgreSQL.
  - Document insertion/removal, vector search queries, and memory cache management.
  - Efficient retrieval and reconstruction of documents with vector data.

**Testing and Usage Example:**

* Provided a comprehensive test script `test/quick-test.js` that:
  - Sets up a test database, inserts sample documents with mock embeddings, and performs vector similarity searches.
  - Demonstrates plugin usage and verifies correct integration with Orama and PostgreSQL.


thatjuan:
Thanks for putting this together. There's real demand for a Postgres/pgvector backend, so I want to be clear about why this version can't merge and what would.

The core problem is that the plugin fights Orama instead of extending it. It hooks `beforeInsert` to strip every property off the document, hooks `afterInsert` to write the row to Postgres and then delete the doc back out of Orama, and hooks `afterSearch` to throw away Orama's result and substitute a raw pgvector query. So the in-memory index is kept permanently empty on purpose, and everything downstream that reads it breaks:

- Text and hybrid search return zero hits every time. `afterSearch` only handles `mode: 'vector'`, and there's nothing left in the index for a term query to match. The README leads with "Hybrid Search," but nothing indexes the text.
- `where` filters, facets, `groupBy`, and `offset` are silently dropped, because the substituted pgvector query doesn't implement any of them. `count` reports the page size, not the match total.
- Restart persistence doesn't work in any mode. `loadAllDocuments` and `loadDocumentIntoMemory` call `orama.insert(doc)`, and that method doesn't exist. Orama's API is functional (`insert(db, doc)`). With `lazyLoad: false` it throws on startup; with `lazyLoad: true` an empty catch swallows the same error and nothing ever loads.

On top of the design, a few concrete bugs would bite the first person who deviates from the exact demo schema:

- `getDocument` hardcodes a column named `embedding_vector`, but columns are created dynamically from the field name. Any schema whose vector field isn't literally called `embedding` (including `sentenceEmbedding` and `imageEmbedding` from your own README) throws `column "embedding_vector" does not exist` on every hit.
- One global `dimension` gets applied to every vector column, and vector detection keys off array length equal to that dimension. The multi-model schema in the README (`vector[384]` plus `vector[512]`) can't work, and a plain `number[]` field that happens to be the right length gets misclassified as a vector.
- `params.vector.property` and `tableName` get interpolated straight into SQL. That's an injection hole for anyone forwarding search params from a request.
- A module-level `removingFromMemory` flag gets toggled around an `await`, and it's shared across every plugin instance in the process. A user `remove()` that lands during an insert gets skipped, so the row never leaves Postgres and keeps showing up in results.

There's also packaging that would need sorting regardless of design: `@orama/orama` sits in `devDependencies` despite being a runtime import, `@tensorflow/tfjs-node` and a couple of others are declared but never imported (that's a ~200MB native build every install pays for nothing), `memoryLimit` is parsed and never read, and the version is `1.0.0` while the workspace is lockstep at `3.2.0` with no entry in the release scripts. I'll skip the rest since the design has to change first.

**What would be acceptable.** Orama already supports swapping the storage and index layers through `components.documentsStore` and `components.index` in `create()`. That's the seam a Postgres backend belongs in. Implement `IDocumentsStore` to keep documents in Postgres, and an `IIndex` (or a vector-specific index) that runs similarity search in pgvector, and you get inserts, removes, filters, facets, and hybrid search routed through the same code paths Orama already tests, with no hook fights, no global flags, and no strip-then-reinsert dance. `plugin-data-persistence` is the closest existing example of replacing a component rather than wrapping the lifecycle.

If you want to take that on, I'm happy to point at the interfaces and review a smaller first cut that just does document storage before adding vector search. A focused component that does one layer correctly is far more valuable than the full feature surface wrapped around hooks. Appreciate the effort here, and I'd like to see it land the right way.

PR #1005, opened 2025-12-14
fix: version name

ilteoood:
Considering [this issue](https://github.com/fastify/website/pull/361), Orama's plugin doesn't seems to work on the fastify's website.
Thanks to the patch proposed in this PR, made by @smith558, we would like to fix this behaviour.


micheleriva:
ETA today

micheleriva:
@ilteoood published

PR #1028, opened 2026-06-30
fix(plugin-match-highlight): highlight every token of a multi-token word (CJK)

mahirhir:
## Problem

In `@orama/plugin-match-highlight`, `recursivePositionInsertion` (the position index used for highlighting) splits a field on a Latin word boundary with `/[\p{L}0-9_'-]+/gimu` and then keeps only the first token of each word:

```ts
;[token] = orama.tokenizer.tokenize(word)
```

CJK scripts have no word spaces, so a whole run is matched as a single word here, and a CJK tokenizer (for example `@orama/tokenizers/mandarin`, which uses `Intl.Segmenter`) splits that run into several tokens. Only the first token gets a recorded position, so every later token is never highlightable even though search matches it.

Example with the mandarin tokenizer:

```
tokenize('我喜欢编程')         -> ['我', '喜欢', '编', '程']
recorded highlight tokens      -> ['我']                  // only the first
searchWithHighlight(db, '程')  -> hit found, positions {} // not highlighted
```

`searchWithHighlight` on the query side already iterates the full token array (`for (const queryToken of queryTokens)`), so the index side keeping only the first token is the inconsistency.

## Fix

Record a position for every token returned by `tokenize(word)`. The recorded span stays the matched word, so single-token words (all Latin cases) are unchanged. After the fix the four tokens above each get a position and `searchWithHighlight(db, '程')` returns the highlight.

## Tests

Added a test that configures a word-granularity `Intl.Segmenter` tokenizer (equivalent to `@orama/tokenizers/mandarin`, no new dependency) and asserts that every token of a CJK run has a recorded position and that a non-leading token is highlightable. Reverting the fix fails the new test on those two assertions. The existing suite is unchanged and passes.

### DIFF
--- a/packages/plugin-match-highlight/src/index.ts
+++ b/packages/plugin-match-highlight/src/index.ts
@@ -71,20 +71,22 @@ async function recursivePositionInsertion<T extends AnyOrama, ResultDocument = T
     let regExResult: RegExpExecArray | null
     while ((regExResult = wordRegEx.exec(text)) !== null) {
       const word = regExResult[0].toLowerCase()
-      const key = `${orama.tokenizer.language}:${word}`
-      let token: string
-      if (orama.tokenizer.normalizationCache.has(key)) {
-        token = orama.tokenizer.normalizationCache.get(key)!
-      } else {
-        ;[token] = orama.tokenizer.tokenize(word)
-        orama.tokenizer.normalizationCache.set(key, token)
-      }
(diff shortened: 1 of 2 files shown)


thatjuan:
Thanks @greymoth-jp. Great diagnosis. No new dependencies, and it fails on both new assertions when I revert the fix. I ran the full plugin suite locally and CI is green.

